Issue #63, opened 2015-06-18
this does not contain special fields like block, elem, mods, elemMods


arikon:
Compiled in production mode template is

```js
function __$b37(__$ctx, __$ref) {
   console.log("this.block", $$block);
   console.log("this.elem", $$elem);
   console.log("this.mods", $$mods);
   console.log("this.elemMods", $$elemMods);
   console.log("this.ctx", __$ctx.ctx);
   return (__$ctx.ctx.items || []).map(function expand__$81(item) {
       console.log("this.block", this.block);
       console.log("this.elem", this.elem);
       console.log("this.mods", this.mods);
       console.log("this.elemMods", this.elemMods);
       console.log("this.ctx", this.ctx);
       if (this.isSimple(item)) {
           item = {
               text: item
           };
       }
       if (item.items) {
           item.elem = "group";
           item.content || (item.content = item.items.map(expand__$81, this));
       } else {
           item.elem = "item";
           if ([ "check", "radio", "radiocheck" ].indexOf(this.mods.type) !== -1) {
               item.elemMods = this.extend({
                   type: "option",
                   checked: this.ctx.val.indexOf(item.val) !== -1 ? "yes" : undefined
               }, item.elemMods);
           } else {
               if (this.mods.type === "navigation") {
                   item.elemMods = this.extend({
                       type: "link"
                   }, item.elemMods);
               }
           }
       }
       return item;
   }, __$ctx);
}
```

The output is:

```
this.block menu
this.elem undefined
this.mods { type: 'check', theme: 'normal', size: 's' }
this.elemMods {}
this.ctx { block: 'menu',
 mods: { type: 'check', theme: 'normal', size: 's' },
 items: [ 'Item 1', 'Item 2', 'Item 3' ],
 val: [] }

this.block undefined
this.elem undefined
this.mods undefined
this.elemMods undefined
this.ctx { block: 'menu',
 mods: { type: 'check', theme: 'normal', size: 's' },
 items: [ 'Item 1', 'Item 2', 'Item 3' ],
 val: [] }
```

The problem is in the second block of console.log()'s. `__$ctx`, that is passed as a context to the `map()` does not contain special fields like `block`, `elem`, `mods`, `elemMods`.

arikon:
Full template source

```js
block('menu')(
    mode('tabindex')(undefined),

    def()(function() {
        var ctx = this.ctx,
            tabindex = this.ctx.tabindex || apply('tabindex');

[...]

PR #65, opened 2015-06-18
compiler: forward-port ContextReplacer from bem-bl

indutny:
Fix: #63

cc @arikon PTAL


indutny:
cc @vkz

arikon:
@indutny In some test there is an error: Maximum call stack size exceeded

indutny:
I think this was landed. @arikon do you have a test case to reproduce the problem?

arikon:
@indutny All the tests are passing now. Templates were rewritten and merged.

indutny:
Cool!

PR #87, opened 2015-08-04
fixes #41 do not throw an error during building if modules is defined…

### LINKED ISSUE #41: An mistake in the wrap method?
I think, instead of 
```js
'(function(g) {\n' +
         '  var __bem_xjst = function(exports' + modulesProvidedDeps + ') {\n' +
         '     ' + code + ';\n' +
         '     return exports;\n' +
         '  }\n' +
         '  var defineAsGlobal = true;\n' +
         '  if(typeof exports === "object") {\n' +
         '    exports["' + exportName + '"] = __bem_xjst({}' + modulesProvidedDeps + ');\n' +
         '    defineAsGlobal = false;\n' +
         '  }\n' +
         '  if(typeof modules === "object") {\n' +
         '    modules.define("' + exportName + '"' + modulesDeps + ',\n' +
         '      function(provide' + modulesProvidedDeps + ') {\n' +
         '        provide(__bem_xjst({}' + modulesProvidedDeps + ')) });\n' +
         '    defineAsGlobal = false;\n' +
         '  }\n' +
         '  defineAsGlobal && (g["' + exportName + '"] = __bem_xjst({}' + modulesProvidedDeps + '));\n' +
         '})(this);'
```

we should use

```js
'(function(g) {\n' +
         '  var __bem_xjst = function(exports' + modulesProvidedDeps + ') {\n' +
         '     ' + code + ';\n' +
         '     return exports;\n' +
         '  }\n' +
         '  var defineAsGlobal = true;\n' +
         '  if(typeof modules === "object") {\n' +
         '    modules.define("' + exportName + '"' + modulesDeps + ',\n' +
         '      function(provide' + modulesProvidedDeps + ') {\n' +
         '        provide(__bem_xjst({}' + modulesProvidedDeps + ')) });\n' +
         '    defineAsGlobal = false;\n' +
         '  }\n' +
         '  else if(typeof exports === "object") {\n' +
         '    exports["' + exportName + '"] = __bem_xjst({}' + modulesProvidedDeps + ');\n' +
         '    defineAsGlobal = false;\n' +
         '  }\n' +
         '  defineAsGlobal && (g["' + exportName + '"] = __bem_xjst({}' + modulesProvidedDeps + '));\n' +
         '})(this);'
```

because now we force users of bemhtml/bemtree to use global variables in node.js context.

jifeon:
#61 for v1.x

It would be great to see it as hot fix in 1.x version before you release v2.0

### DIFF
--- a/lib/bemhtml/compiler.js
+++ b/lib/bemhtml/compiler.js
@@ -726,14 +726,26 @@ Compiler.prototype.wrap = function wrap(code) {
   if (!this.options.wrap)
     return code;
 
-  var opts = this.options;
-  var exportName = opts.exportName || 'BEMHTML';
-  var deps = opts.modulesDeps;
-  var modulesDeps = deps ? ', ' + JSON.stringify(Object.keys(deps)) : '';
-  var modulesProvidedDeps =  deps ? ', ' + Object.keys(deps).map(function(module) {
-    var providedName = deps[module];
-    return providedName === true ? module : providedName;
-  }).join(', ') : '';
+  var opts = this.options,
+    exportName = opts.exportName || 'BEMHTML',
+    deps = opts.modulesDeps,
+    modulesDeps,
+    modulesProvidedDeps,
+    allGlobalsDefinedCheck;
+
+  if (!deps) {
+    modulesDeps = '';
+    modulesProvidedDeps = '';
+    allGlobalsDefinedCheck = 'true';
+  } else {
+    modulesDeps = ', ' + JSON.stringify(Object.keys(deps));
+    var modulesProvidedDepsNames = Object.keys(deps).map(function (module) {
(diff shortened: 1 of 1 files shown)


tadatuta:
JFYI, there's already `4.0.1` and we have `1.x` and `4.x` branches in support currently

tadatuta:
looks like we don't need it anymore after https://github.com/enb-bem/enb-bemxjst/pull/83/files ?

apsavin:
Can I check it with the latest bem-core?

tadatuta:
@apsavin yes, latest commit to `bem-core` has the version with this change of `enb-bemxjst`

apsavin:
Yes, looks like we don't need it anymore.

Issue #113, opened 2015-10-30
Error only exists at first BEMHTML.apply


sbmaxx:
How to reproduce:

```
git clone https://github.yandex-team.ru/sbmaxx/bemhtml-bug.git
cd bemhtml-bug
npm i
npm test
```

On first run i have: `[TypeError: Cannot set property 'onmousedown' of undefined]`
Second — all is fine.

Templates from islands 4.0.1, "compiled" with enb-bemxjst 4.0.0

sbmaxx:
@veged @arikon

tadatuta:
@sbmaxx are you sure it's a bug?

seems like you've got following situation:

```js
block('b1').attrs()(function() {
    var prevAttrs = applyNext(); // undefined as it's just a first template for b1
    prevAttrs.someKey = 'blah'; // can't set property 'someKey' of undefined
});
```
so all you need is just to add fallback for `fiji/images/blocks-deskpad/snippet/_type/snippet_type_series.bemhtml.js` like this:

```js
/* ... */
elem('link')(
        attrs()(function() {
            var attrs = applyNext() || {}; // here it is
            this.ctx.counter && (attrs.onmousedown = this.ctx.counter);
            return attrs;
        })
/* ... */
```

sbmaxx:
@tadatuta i know that template are ugly and incorrect. The problem is "why it reproduces only once or twice"

tadatuta:
It's because script failed at https://github.yandex-team.ru/sbmaxx/bemhtml-bug/blob/master/bemhtml.js#L1278 and as so [these lines](https://github.yandex-team.ru/sbmaxx/bemhtml-bug/blob/master/bemhtml.js#L1282-L1284) were never reached so on next `BEMHTML.apply()` call the template is considered already visited.

Still I'm not sure if it's a bug or a feature.

sbmaxx:
I think it should fail always. Consider how to debug errors if they disappear after refresh? Inspired by migration to islands 4.0.1 ;)

tadatuta:
adding `try catch` around `out = template.body.call(context);` will make it happen but it can be expensive in terms of performance so let's wait for @indutny to take a look.

sbmaxx:
We can almost level performance with extracted `try/catch`:
```js
Match.prototype.tryCatch = function(fn, ctx) {
    try {
        return fn.call(ctx);
    } catch(e) {
        return 'SOME_ERROR';
    }
};
```

Results:
```
[...]

PR #114, opened 2015-11-01
try/catch apply

sbmaxx:
Fixes for https://github.com/bem/bem-xjst/issues/113

### DIFF
--- a/lib/bemhtml/runtime/match.js
+++ b/lib/bemhtml/runtime/match.js
@@ -167,12 +167,12 @@ Match.prototype.push = function push(template) {
 };
 
 Match.prototype.tryCatch = function(fn, ctx) {
-    try {
-        return fn.call(ctx);
-    } catch(e) {
-        console.error(e);
-        return '';
-    }
+  try {
+    return fn.call(ctx);
+  } catch(e) {
+    console.error(e);
+    return '';
+  }
 };
 
 Match.prototype.exec = function exec(context) {
@@ -216,11 +216,10 @@ Match.prototype.exec = function exec(context) {
   this.bemhtml.match = this;
 
   var out;
-  if (typeof template.body === 'function') {
+  if (typeof template.body === 'function')
     out = this.tryCatch(template.body,  context);
-  } else {
+  else
     out = template.body;
-  }
 
   this.mask[bitIndex] = oldMask;
   this.bemhtml.match = oldMatch;
--- a/lib/bemhtml/runtime/match.js
+++ b/lib/bemhtml/runtime/match.js
@@ -227,15 +227,13 @@ Match.prototype.exec = function exec(context) {
   this.bemhtml.match = oldMatch;
   this.restoreDepth(save);
 
-  var e;
-  if (this.thrownError !== null) {
-    e = this.thrownError;
+  var e = this.thrownError;
+  if (e !== null) {
     this.thrownError = null;
     throw e;
   }
 
   return out;
-
 };
 
 Match.prototype.checkDepth = function checkDepth() {
--- a/lib/bemhtml/runtime/match.js
+++ b/lib/bemhtml/runtime/match.js
@@ -170,8 +170,7 @@ Match.prototype.tryCatch = function(fn, ctx) {
   try {
     return fn.call(ctx);
   } catch (e) {
-    console.error(e);
-    return '';
+    return e;
   }
 };
 
@@ -224,6 +223,9 @@ Match.prototype.exec = function exec(context) {
   this.mask[bitIndex] = oldMask;
   this.bemhtml.match = oldMatch;
   this.restoreDepth(save);
+
+  if (out instanceof Error)
+    throw out;
 
   return out;
 };
--- a/lib/bemhtml/runtime/match.js
+++ b/lib/bemhtml/runtime/match.js
@@ -132,6 +132,8 @@ function Match(entity) {
 
   this.count = 0;
   this.depth = -1;
+
+  this.thrownError = null;
 }
 exports.Match = Match;
 
@@ -168,9 +170,10 @@ Match.prototype.push = function push(template) {
 
 Match.prototype.tryCatch = function(fn, ctx) {
   try {
+    this.thrownError = null;
     return fn.call(ctx);
   } catch (e) {
-    return e;
+    this.thrownError = e;
   }
 };
 
@@ -216,7 +219,7 @@ Match.prototype.exec = function exec(context) {
 
   var out;
   if (typeof template.body === 'function')
-    out = this.tryCatch(template.body,  context);
+    out = this.tryCatch(template.body, context);
   else
     out = template.body;
 
@@ -224,10 +227,15 @@ Match.prototype.exec = function exec(context) {
   this.bemhtml.match = oldMatch;
   this.restoreDepth(save);
 
-  if (out instanceof Error)
-    throw out;
+   var e;
+   if (this.thrownError !== null) {
+     e = this.thrownError;
+     this.thrownError = null;
+     throw e;
+   }
 
   return out;
+
 };
 
 Match.prototype.checkDepth = function checkDepth() {
--- a/lib/bemhtml/runtime/match.js
+++ b/lib/bemhtml/runtime/match.js
@@ -166,6 +166,15 @@ Match.prototype.push = function push(template) {
   this.maskSize = this.mask.length;
 };
 
+Match.prototype.tryCatch = function(fn, ctx) {
+    try {
+        return fn.call(ctx);
+    } catch(e) {
+        console.error(e);
+        return '';
+    }
+};
+
 Match.prototype.exec = function exec(context) {
   var save = this.checkDepth();
 
@@ -207,10 +216,11 @@ Match.prototype.exec = function exec(context) {
   this.bemhtml.match = this;
 
   var out;
-  if (typeof template.body === 'function')
-    out = template.body.call(context);
-  else
+  if (typeof template.body === 'function') {
+    out = this.tryCatch(template.body,  context);
+  } else {
     out = template.body;
+  }
 
   this.mask[bitIndex] = oldMask;
   this.bemhtml.match = oldMatch;
--- a/lib/bemhtml/runtime/match.js
+++ b/lib/bemhtml/runtime/match.js
@@ -169,7 +169,7 @@ Match.prototype.push = function push(template) {
 Match.prototype.tryCatch = function(fn, ctx) {
   try {
     return fn.call(ctx);
-  } catch(e) {
+  } catch (e) {
     console.error(e);
     return '';
   }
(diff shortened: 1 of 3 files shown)


indutny:
I think it is very dangerous change. It may obscure things running in production, without giving our any kind of introspection. IMO, we should find a better way for it.

sbmaxx:
@indutny any ideas?

indutny:
I think we should save the error here: https://github.com/bem/bem-xjst/blob/master/lib/bemhtml/runtime/match.js#L211 and re-throw it after restoring all values: https://github.com/bem/bem-xjst/blob/master/lib/bemhtml/runtime/match.js#L217

sbmaxx:
@indutny like this https://github.com/bem/bem-xjst/commit/5b10fb3fa4da600ece866db324d0e5a6cf1aec98 ?

indutny:
@sbmaxx this is not as great either. Why not wrap `template.body.call()` into try-catch in-place and store exception to the local variable there. If that local variable is non-empty at the end of the `exec` - throw exception.

Does it sound right?

sbmaxx:
Just because of perf:

**Inined** `try/catch`:
```
sbmaxx@sbmaxx-mba ~/Develop/bem-xjst/bench $ make -B -C ../ && node run --compare
make: Circular lib/bemhtml/runtime/bundle.js <- lib/bemhtml/runtime/bundle.js dependency dropped.
./node_modules/.bin/browserify --standalone BEMHTML lib/bemhtml/runtime/index.js -o lib/bemhtml/runtime/bundle.js
render:basic:next x 240,096 ops/sec ±1.05% (86 runs sampled)
render:basic:prev x 299,737 ops/sec ±1.22% (85 runs sampled)
render:islands:next x 473 ops/sec ±1.34% (84 runs sampled)
render:islands:prev x 617 ops/sec ±1.38% (84 runs sampled)
render:showcase:next x 999 ops/sec ±1.28% (83 runs sampled)
render:showcase:prev x 1,212 ops/sec ±2.72% (83 runs sampled)
# one more try
render:basic:next x 244,273 ops/sec ±1.11% (86 runs sampled)
render:basic:prev x 296,970 ops/sec ±1.06% (87 runs sampled)
render:islands:next x 469 ops/sec ±1.54% (84 runs sampled)
render:islands:prev x 590 ops/sec ±2.09% (84 runs sampled)
render:showcase:next x 819 ops/sec ±5.02% (82 runs sampled)
render:showcase:prev x 1,094 ops/sec ±2.94% (81 runs sampled)
```

**Prototype.tryCatch**:
```
sbmaxx@sbmaxx-mba ~/Develop/bem-xjst/bench $ make -B -C ../ && node run --compare
make: Circular lib/bemhtml/runtime/bundle.js <- lib/bemhtml/runtime/bundle.js dependency dropped.
./node_modules/.bin/browserify --standalone BEMHTML lib/bemhtml/runtime/index.js -o lib/bemhtml/runtime/bundle.js
render:basic:next x 292,620 ops/sec ±1.11% (87 runs sampled)
render:basic:prev x 302,492 ops/sec ±1.15% (86 runs sampled)
render:islands:next x 554 ops/sec ±1.56% (85 runs sampled)
render:islands:prev x 572 ops/sec ±2.22% (85 runs sampled)
render:showcase:next x 1,051 ops/sec ±2.10% (85 runs sampled)
render:showcase:prev x 1,166 ops/sec ±2.25% (82 runs sampled)
```

sbmaxx:
@indutny seems like inlined try/catch has bigger performance penalty, can you check this?

indutny:
@sbmaxx I will definitely do, but later today.

sbmaxx:
ok, thx! :)

indutny:
Btw, I have strong doubts that this thing is going to work properly even after this fix. Throwing exceptions will leave dangling changes made by `local(...)`. Perhaps we are trying to fix the thing that should not be working by-design? cc @veged

veged:
let's discuss it offline at our regular meeting...

sbmaxx:
Any news?

zxqfox:
> seems like inlined try/catch has bigger performance penalty, can you check this?

@sbmaxx Sure it is. Any code block with try-catch can't be optimized by v8. Why not to make universal `tryCatch` and use it everywhere in project?

```
function tryCatch(fn, cb) {
  try { return fn(); } catch(e) { cb(e); }
}
//...
  var err;
  out = tryCatch(template.body.bind(context), function (e) { err = e; });
// ...
if (err) throw err;
return out;
```

lib/bemhtml/runtime/match.js
```diff
@@ -208,7 +216,7 @@ Match.prototype.exec = function exec(context) {
 
   var out;
   if (typeof template.body === 'function')
-    out = template.body.call(context);
+    out = this.tryCatch(template.body,  context);
```
zxqfox: double space before context

lib/bemhtml/runtime/match.js
```diff
@@ -166,6 +166,14 @@ Match.prototype.push = function push(template) {
   this.maskSize = this.mask.length;
 };
 
+Match.prototype.tryCatch = function(fn, ctx) {
+  try {
+    return fn.call(ctx);
+  } catch (e) {
+    return e;
```
indutny: Let's create a property called `.thrownError` in constructor, and initialize it to `null`.

indutny: Instead of `return e`, we should now do `this.thrownError = e`.

lib/bemhtml/runtime/match.js
```diff
@@ -166,6 +166,14 @@ Match.prototype.push = function push(template) {
   this.maskSize = this.mask.length;
 };
 
+Match.prototype.tryCatch = function(fn, ctx) {
+  try {
+    return fn.call(ctx);
```
indutny: And before the call it should be set to `null` again.

lib/bemhtml/runtime/match.js
```diff
@@ -216,6 +224,9 @@ Match.prototype.exec = function exec(context) {
   this.bemhtml.match = oldMatch;
   this.restoreDepth(save);
 
+  if (out instanceof Error)
```
indutny: This should check if `.thrownError` is not `null`, load it, cleanup the field (set to `null`), and throw  it.

lib/bemhtml/runtime/match.js
sbmaxx: :gem: eyes ;)

i think we need jscs here

indutny: We have it.

sbmaxx:
@indutny fixed. Also added some tests about this fix

cc @zxqfox, @veged

lib/bemhtml/runtime/match.js
```diff
@@ -216,7 +227,15 @@ Match.prototype.exec = function exec(context) {
...
 
+  var e;
+  if (this.thrownError !== null) {
+    e = this.thrownError;
+    this.thrownError = null;
+    throw e;
+  }
+
   return out;
+
```
zxqfox: Quitely fine PR imho, except this line. ;-)

lib/bemhtml/runtime/match.js
```diff
@@ -216,7 +227,15 @@ Match.prototype.exec = function exec(context) {
   this.bemhtml.match = oldMatch;
   this.restoreDepth(save);
 
+  var e;
+  if (this.thrownError !== null) {
+    e = this.thrownError;
```
zxqfox: NonBlocker: btw, you can reduce 1 line of code with moving this to `var`. And also you can replace `this.thrown` in `if` with just `e`.

indutny: Yeah, please move `var e` here.

lib/bemhtml/runtime/match.js
indutny: Agree :)

test/runtime-test.js
```diff
@@ -880,4 +880,44 @@ describe('BEMHTML compiler/Runtime', function() {
...
+    it('should throw errors with many applies', function() {
+      assert.throws(function() {
+        template.apply(bemjson);
+      });
+      assert.throws(function() {
+        template.apply(bemjson);
+      });
+    });
+  });
+
```
indutny: Excessive newline.

indutny:
Do benchmarks still look good?

lib/bemhtml/runtime/match.js
sbmaxx: done

sbmaxx:
> Do benchmarks still look good?

```bash
sbmaxx@sbmaxx-mba ~/Develop/bem-xjst/bench $ node run.js --compare
render:basic:next x 176,977 ops/sec ±8.47% (78 runs sampled)
render:basic:prev x 301,332 ops/sec ±1.18% (86 runs sampled)
render:islands:next x 607 ops/sec ±1.09% (86 runs sampled)
render:islands:prev x 580 ops/sec ±1.41% (86 runs sampled)
render:showcase:next x 1,146 ops/sec ±0.71% (86 runs sampled)
render:showcase:prev x 1,178 ops/sec ±2.34% (84 runs sampled)

sbmaxx@sbmaxx-mba ~/Develop/bem-xjst/bench $ node run.js --compare
render:basic:next x 294,854 ops/sec ±1.00% (87 runs sampled)
render:basic:prev x 297,137 ops/sec ±1.05% (87 runs sampled)
render:islands:next x 522 ops/sec ±2.91% (81 runs sampled)
render:islands:prev x 596 ops/sec ±1.27% (86 runs sampled)
render:showcase:next x 1,158 ops/sec ±1.30% (85 runs sampled)
render:showcase:prev x 1,182 ops/sec ±2.63% (84 runs sampled)

sbmaxx@sbmaxx-mba ~/Develop/bem-xjst/bench $ node run.js --compare
render:basic:next x 279,900 ops/sec ±1.06% (86 runs sampled)
render:basic:prev x 310,417 ops/sec ±0.42% (88 runs sampled)
render:islands:next x 593 ops/sec ±1.42% (85 runs sampled)
render:islands:prev x 607 ops/sec ±1.61% (87 runs sampled)
render:showcase:next x 1,108 ops/sec ±1.30% (85 runs sampled)
render:showcase:prev x 1,233 ops/sec ±2.03% (85 runs sampled)
```

indutny:
Landed with minor fixes in 1ab21c35236b861bae8ffd0d79c3ef1713982d69. Thank you!

Btw, stable benchmark results could be obtained with `--min-samples` flag:

```bash
$ node bench/run.js --min-samples=1000 --compare
render:basic:next x 877,088 ops/sec ±0.26% (1038 runs sampled)
render:basic:prev x 935,638 ops/sec ±0.19% (1038 runs sampled)
render:islands:next x 1,412 ops/sec ±0.28% (1035 runs sampled)
render:islands:prev x 1,458 ops/sec ±0.32% (1036 runs sampled)
render:showcase:next x 2,598 ops/sec ±0.34% (1035 runs sampled)
render:showcase:prev x 2,869 ops/sec ±0.25% (1038 runs sampled)
```

zxqfox:
Great! Thanks

PR #115, opened 2015-11-01
Editorconfig

sbmaxx:
Add support for `.editorconfig`. It's important, because "by default" only small amount of people use 2 spaces in JS files ;)


indutny:
That depends :) But I'm fine with it. Thanks!

Issue #119, opened 2015-11-06
generateId is drunk


sbmaxx:
http://bem.github.io/bem-xjst/

*bemjson*
```
({
    block: 'block',
    content: [1,2,3,4,5].map(function() {
        return { block: 'ololo' };
    })
});
```

*bemhtml*
```
block('ololo').content()(function() {
    return this.generateId()
});
```

All IDs are the same:
```
<div class="block">
    <div class="ololo">uniq144683410696013</div>
    <div class="ololo">uniq144683410696013</div>
    <div class="ololo">uniq144683410696013</div>
    <div class="ololo">uniq144683410696013</div>
    <div class="ololo">uniq144683410696013</div>
</div>
```

And yes, this is also critical as https://github.com/bem/bem-xjst/pull/114

zxqfox:
This generated id is stored in ._uniq property of this class: https://github.com/bem/bem-xjst/blob/master/lib/bemhtml/runtime/index.js#L256

I'm not sure why, but there is only one instance of that. This behaviour is right from v2.0.0.

Why you can't use .identify? ;-)

sbmaxx:
> Why you can't use .identify? ;-)

I can, but i don't want to extend all input* islands templates ;)

zxqfox:
Not drunk anymore. ;-)

PR #120, opened 2015-11-07
fix generateId

sbmaxx:
#119

### DIFF
--- a/lib/bemhtml/runtime/context.js
+++ b/lib/bemhtml/runtime/context.js
@@ -17,8 +17,6 @@ function Context(bemhtml) {
   this._listLength = 0;
   this._notNewList = false;
 
-  this._uniq = null;
-
   // Used in `OnceMatch` check to detect context change
   this._onceRef = {};
 }
@@ -48,9 +46,7 @@ Context.prototype.isLast = function isLast() {
 };
 
 Context.prototype.generateId = function generateId() {
-  if (this._uniq === null)
-    this._uniq = utils.getUniq();
-  return this._uniq;
+  return utils.identify(this.ctx);
 };
 
 Context.prototype.reapply = function reapply(ctx) {
--- a/test/runtime-test.js
+++ b/test/runtime-test.js
@@ -625,6 +625,58 @@ describe('BEMHTML compiler/Runtime', function() {
          '<div class="b1">first</div>' +
          '<div class="b1 b1_position_last">last</div>');
     });
+
+    describe('generateId', function() {
+      var template;
+
+      beforeEach(function() {
+        template = bemxjst.compile(function() {
+          block('b1')(
+            tag()(''),
+            content()(function() {
+              return this.generateId();
+            })
+          )
+        });
+      });
+
+      it('starts with uniq', function() {
+        var str = template.apply({ block: 'b1' });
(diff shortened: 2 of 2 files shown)


sbmaxx:
@veged @indutny take a look, please

lib/bemhtml/runtime/context.js
```diff
@@ -48,9 +46,7 @@ Context.prototype.isLast = function isLast() {
 };
 
 Context.prototype.generateId = function generateId() {
-  if (this._uniq === null)
-    this._uniq = utils.getUniq();
-  return this._uniq;
+  return utils.getUniq();
```
indutny: To be compatible with old bem-xjst, it should actually generate id based on current context. Is it ok to generate random here, @veged ?

zxqfox: +1 for `this.identify(this.ctx)` analogue.

zxqfox: @sbmaxx try this:
```diff
$ git diff
diff --git a/lib/bemhtml/runtime/context.js b/lib/bemhtml/runtime/context.js
index 3e275ec..d209f3e 100644
--- a/lib/bemhtml/runtime/context.js
+++ b/lib/bemhtml/runtime/context.js
@@ -46,7 +46,7 @@ Context.prototype.isLast = function isLast() {
 };

 Context.prototype.generateId = function generateId() {
-  return utils.getUniq();
+  return utils.identify(this.ctx);
 };

 Context.prototype.reapply = function reapply(ctx) {
```

veged: yes, it must be same for same node

sbmaxx:
@veged @zxqfox fixed

test/runtime-test.js
```diff
@@ -625,6 +625,48 @@ describe('BEMHTML compiler/Runtime', function() {
...
+            tag()(''),
+            content()(function() {
+              return this.generateId();
+            })
+          )
+        })
+      });
+
+      it('starts with uniq', function() {
+
```
zxqfox: Here you have a spacer

test/runtime-test.js
```diff
@@ -625,6 +625,48 @@ describe('BEMHTML compiler/Runtime', function() {
...
+        var str = template.apply({ block: 'b1' });
+        assert(str.indexOf('uniq') === 0);
+
+      });
+
+      it('should be unique in different applies', function() {
+        assert.notEqual(
```
zxqfox: And here you don't

test/runtime-test.js
```diff
@@ -625,6 +625,48 @@ describe('BEMHTML compiler/Runtime', function() {
...
+          template.apply({ block: 'b1' }),
+          template.apply({ block: 'b1' })
+        );
+      });
+
+      it('should be unique in one apply', function() {
+
```
zxqfox: And again ;-)

zxqfox:
I think we missing one more test case:
```
block('b')(
  tag()(''),
  content()(function(){
    return [this.generateId(), sep, this.generateId()];
  })
);
// and
var res = t.apply({block: 'b'}).split(sep);
assert.equal(res[0], res[1]);
```

sbmaxx:
@zxqfox thx for review ;)

PR #125, opened 2015-11-18
Support attrs values as an object

### LINKED ISSUE #124: Fails if attr is object
**BEMHTML:**

```js
block('block').def()(function() {
    var ctx = this.ctx;

    return applyCtx({
        attrs: {
            'aria-valuenow': ctx.value,
        })
    });
});
```

**BEMJSON:**

```js
{
    block: 'block',
    value: {
        block: 'i-bem',
        elem: 'i18n',
        keyset: 'keyset',
        key: 'key'
    }
}
```

Expected:

```html
<div aria-valuenow="[object Object]"></div>
```

But now `BEMHTML.apply()` fails with error:

```
TypeError: undefined is not a function
```

Stack trace points to this [code](
https://github.com/bem/bem-xjst/blob/cf8aea838292374c14b026e42de5647de63a8658/lib/bemhtml/runtime/index.js#L514).

tadatuta:
Closes #124

WARNING! Looks like bem-xjst 1.x results with `some-attr=[object Object]` for `attrs()({ 'some-attr': { block: 'b1' } })` (see expected section of #124) which I think is also wrong.

cc @indutny @veged @blond

### DIFF
--- a/lib/bemhtml/runtime/index.js
+++ b/lib/bemhtml/runtime/index.js
@@ -507,11 +507,10 @@ BEMHTML.prototype.renderClose = function renderClose(prefix,
       if (attr === undefined)
         continue;
 
-      // TODO(indutny): support `this.reapply()`
       out += ' ' + name + '="' +
         utils.attrEscape(utils.isSimple(attr) ?
                          attr :
-                         this.reapply(attr)) +
+                         this.context.reapply(attr)) +
                          '"';
     }
   }
--- a/test/runtime-test.js
+++ b/test/runtime-test.js
@@ -209,6 +209,16 @@ describe('BEMHTML compiler/Runtime', function() {
     }, { block: 'b1' }, '<div class="b1"><div class="b2">ok</div></div>');
   });
 
+  it('should support objects as attrs values', function() {
+    test(function() {
+      block('b1').attrs()(function() {
+        return { prop1: { block: 'b2' } };
+      });
+
+      block('b2').replace()('hello');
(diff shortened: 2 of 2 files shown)


indutny:
Any performance impact?

tadatuta:
```
node run.js --compare
render:basic:next x 624,194 ops/sec ±2.37% (83 runs sampled)
render:basic:prev x 639,001 ops/sec ±1.86% (87 runs sampled)
render:islands:next x 911 ops/sec ±4.47% (83 runs sampled)
render:islands:prev x 964 ops/sec ±2.73% (85 runs sampled)
render:showcase:next x 1,831 ops/sec ±2.70% (84 runs sampled)
render:showcase:prev x 1,701 ops/sec ±3.28% (83 runs sampled)


node run.js --compare
render:basic:next x 658,996 ops/sec ±2.48% (85 runs sampled)
render:basic:prev x 650,483 ops/sec ±2.25% (87 runs sampled)
render:islands:next x 921 ops/sec ±3.78% (83 runs sampled)
render:islands:prev x 928 ops/sec ±2.67% (82 runs sampled)
render:showcase:next x 1,719 ops/sec ±2.26% (85 runs sampled)
render:showcase:prev x 1,721 ops/sec ±5.97% (83 runs sampled)
```

But I'm not quite sure if current benchs are in any way affected by the update as there's no templates with attrs with object values so changed code is never used.

indutny:
Ok, LGTM

PR #129, opened 2015-11-26
Context: Remove unused BEM field

tadatuta:
Can't find any usage of this field. Is it safe to remove?

cc @veged @indutny


indutny:
I have no idea, put it here just in case.

tadatuta:
@narqo do you have any considerations about `BEM` field in BEMHTML context?

narqo:
I think it was used only to store `INTERNAL` stuff. Not sure we have any usecases where it might be used.

PR #132, opened 2015-11-26
Introduce BEMTREE support

tadatuta:
This is breaking change.

cc @veged @indutny @vithar


lib/bemxjst/entity.js
```diff
@@ -56,6 +40,7 @@ Entity.prototype.initModes = function initModes(templates) {
 
     for (var j = template.predicates.length - 1; j >= 0; j--) {
       var pred = template.predicates[j];
+
```
indutny: Irrelevant change.

lib/bemxjst/entity.js
```diff
@@ -35,20 +25,14 @@ function Entity(bemhtml, block, elem, templates) {
   this.init(block, elem);
   this.initModes(templates);
 }
+
```
indutny: Irrelevant change.

lib/compiler.js
```diff
@@ -0,0 +1,69 @@
+var vm = require('vm');
+
+function Compiler(options) {
+  this.options = options || {};
+}
+
+module.exports = function(runtime, sourceBundle) {
```
indutny: Please name the function here.

lib/compiler.js
```diff
@@ -0,0 +1,69 @@
...
+  runtime.source = sourceBundle;
+  Compiler.prototype.runtime = runtime;
```
indutny: So calling this two times will do something strange, right? Perhaps it should return a new `Compiler` constructor each time?

test/bemhtml/runtime-test.js
```diff
@@ -338,12 +338,12 @@ describe('BEMHTML compiler/Runtime', function() {
       }, {
         block: 'b1',
         elemMods: {
-          a: 'yes'
+          a: 'no'
```
indutny: What's the point of this change?

index.js
```diff
@@ -0,0 +1,26 @@
+var fs = require('fs');
+
+function getEngine(engineName) {
+  var runtime = require('./lib/' + engineName);
+  var sourceBundle = fs.readFileSync(require.resolve('./lib/' + engineName + '/bundle'), 'utf8');
```
indutny: Does it fit into 80 columns?

index.js
```diff
@@ -0,0 +1,26 @@
...
+  var pathToCompiler = require.resolve('./lib/compiler');
+  var Compiler = require(pathToCompiler)(runtime, sourceBundle);
```
indutny: Why not just use `require('./lib/compiler')`? And require it outside of the `getEngine` function?

index.js
```diff
@@ -0,0 +1,26 @@
...
+
+  // we need two independent instances of Compiler for bemhtml and bemtree
+  delete require.cache[pathToCompiler];
```
indutny: This should not be necessary. See my comments below.

lib/bemtree/entity.js
```diff
@@ -0,0 +1,25 @@
+var Match = require('../bemxjst/match').Match;
+var Entity = require('../bemxjst/entity').Entity;
+
+exports.Entity = Entity;
```
indutny: Should use inheritance instead of monkey-patching existing constructor.

lib/bemxjst/entity.js
```diff
@@ -142,6 +92,25 @@ Entity.prototype.prepend = function prepend(other) {
   }
 };
 
+function contentMode() {
+  return this.ctx.content;
+}
+
+Entity.prototype.setDefaults = function setDefaults() {
```
indutny: Any point in shuffling this around?

lib/bemxjst/entity.js
```diff
@@ -35,20 +25,14 @@ function Entity(bemhtml, block, elem, templates) {
...
 
 Entity.prototype.init = function init(block, elem) {
   this.block = block;
   this.elem = elem;
-
-  // Class for jsParams
-  this.jsClass = this.bemhtml.classBuilder.build(this.block, this.elem);
 };
 
-function contentMode() {
```
indutny: Any point in shuffling this around?

lib/bemxjst/index.js
```diff
@@ -38,14 +37,15 @@ function BEMHTML(options) {
   this.oninit = null;
 
   // Initialize default entity (no block/elem match)
-  this.defaultEnt = new Entity(this, '', '', []);
-  this.defaultElemEnt = new Entity(this, '', '', []);
+  this.defaultEnt = new BEMXJST.prototype.Entity(this, '', '', []);
```
indutny: should use `new this.Entity(...)`

lib/bemxjst/index.js
```diff
@@ -20,8 +19,8 @@ function BEMHTML(options) {
   this.match = null;
 
   // Create new Context constructor for overriding prototype
-  this.contextConstructor = function ContextChild(bemhtml) {
-    Context.call(this, bemhtml);
+  this.contextConstructor = function ContextChild(bemxjst) {
+    Context.call(this, bemxjst);
```
indutny: I think there might be reason to inherit it from `this.Context`, as in future we may like to have two different kinds of `Context` for BEMTREE and BEMHTML.

lib/bemxjst/index.js
```diff
@@ -113,14 +113,17 @@ BEMHTML.prototype.compile = function compile(code) {
...
-  var args = BEMHTML.locals;
+  var args = BEMXJST.prototype.locals;
+  if (typeof code === 'function') {
+    var start = out.match(/^function\s*\(([^{]+)\)\s*{/);
 
-  // Reuse function if it already has right arguments
-  if (typeof code === 'function' && code.length === args.length)
-    return code;
+    // Reuse function if it already has right arguments
+    if (start !== null && start[1].split(/,\s*/).length === args.length)
```
indutny: I think previous way was better.

lib/bemxjst/index.js
```diff
@@ -185,7 +188,8 @@ BEMHTML.prototype.transformEntities = function transformEntities(entities) {
     if (elem === '*')
       wildcardElems.push(block);
 
-    entities[key] = new Entity(this, block, elem, entities[key]);
+    entities[key] = new BEMXJST.prototype.Entity(
```
indutny: `this.Entity`

lib/bemhtml/entity.js
```diff
@@ -0,0 +1,84 @@
...
+  this.js = new Match(this);
+  this.mix = new Match(this);
+  this.bem = new Match(this);
+  this.cls = new Match(this);
+
+  BemxjstEntity.apply(this, arguments);
+}
+
+/* jshint unused : false */
+inherits(Entity, BemxjstEntity);
```
indutny: What's unused here? Result?

voischev:
Ee!

test/bemhtml/runtime-test.js
tadatuta: to differentiate mod val from content

indutny: Mmm... ok

lib/bemxjst/index.js
tadatuta: Let's postpone it until real need?

tadatuta:
@indutny 
All notes except for premature inheritance from `Context` were fixed, see last 4 commits.

Still need your help to fix falling test with `reapply()` in BEMTREE. And please give us some clue about proper way to implement async version of BEMTREE.

lib/bemxjst/index.js
indutny: Ok, I don't mind.

lib/bemxjst/index.js
```diff
@@ -113,14 +113,15 @@ BEMHTML.prototype.compile = function compile(code) {
...
 
-  var args = BEMHTML.locals;
-
-  // Reuse function if it already has right arguments
-  if (typeof code === 'function' && code.length === args.length)
-    return code;
+  var args = BEMXJST.prototype.locals;
+  if (typeof code === 'function') {
+    // Reuse function if it already has right arguments
+    if (typeof code === 'function' && code.length === args.length)
```
indutny: Two `typeof`? Kind of pointless?

lib/bemxjst/tree.js
```diff
@@ -302,12 +302,14 @@ Tree.prototype.match = function match() {
...
 Tree.prototype.once = function once() {
-  if (arguments.length) throw new Error('Predicate should not have arguments');
+  if (arguments.length)
+    throw new Error('Predicate once() should not have arguments');
   return this.match(new OnceMatch());
 };
 
 Tree.prototype.applyMode = function applyMode(args, mode) {
-  if (args.length) throw new Error('Predicate should not have arguments');
+  if (args.length) throw new Error('Predicate should not have arguments but ' +
```
indutny: Let's add `{}` braces here and put `throw` on a new line between them.

package.json
```diff
@@ -1,14 +1,14 @@
...
   "bin": {
     "bem-xjst": "bin/bem-xjst"
   },
   "scripts": {
     "prepublish": "make",
-    "bundle": "echo \"var fs = require('fs'), xjst = require('bem-xjst'); fs.writeFileSync('bundle.js', xjst.generate(''));\" | node",
-    "test": "make && make lint && mocha --reporter=spec test/*-test.js"
+    "test-bemhtml": "mocha --reporter=spec test/bemhtml/*-test.js",
+    "test-bemtree": "mocha --reporter=spec test/bemtree/*-test.js",
+    "test": "make && make lint && npm run test-bemhtml && npm run test-bemtree"
```
zxqfox: Why to use npm scripts and make in parallel? ;-(

test/bemtree/runtime-test.js
```diff
@@ -0,0 +1,739 @@
...
+      });
+
+      assert.deepEqual(template.apply({ block: 'b1' }), 'new');
+      assert.deepEqual(
+        template.apply({ block: 'b2' }), { block: 'b2', content: 'ok' }
+      );
+    });
+  });
+
+  it('should work with empty input', function() {
```
zxqfox: Why not to group these as basics?

zxqfox:
@tadatuta Great job! Would be very helpful if we start group tests as sections in further specification.

tadatuta:
@indutny :up: 

I also got rid of Makefile as @zxqfox suggested.

> Why not to group these as basics?

Let's make tests improvements in separate PR?

zxqfox:
> Let's make tests improvements in separate PR?

@tadatuta Yup! Sorry.

index.js
```diff
@@ -0,0 +1,25 @@
...
+      return new Compiler(options, runtime).generate(code, options);
+    },
+    compile: function(code, options) {
+      return new Compiler(options, runtime).compile(code);
+    }
+  }
+}
+
+module.exports = {
+  bemtree: function() { return getEngine('bemtree'); },
```
indutny: What do you think about making this a getter with some sort of caching instead?

tadatuta:
@indutny @veged :up:

veged:
LGTM

voischev:
:tada: :tada: :tada: в этом году выпустите? ;)

lib/bemxjst/entity.js
```diff
@@ -150,32 +102,17 @@ Entity.prototype.run = function run(context) {
...
-                             attrs,
-                             content);
+Entity.prototype.setDefaults = function setDefaults() {
+  // Default .content() template for applyNext()
+  if (this.content.count !== 0)
+    this.content.push(new Template([], contentMode));
+
+  // .def() default
+  if (this.def.count !== 0) {
+    this.canFlush = this.options.flush || false;
```
indutny: This is still set here, but the property is declared only on a child class. What do you think about moving the property back to this file?

lib/bemhtml/index.js
```diff
@@ -0,0 +1,316 @@
+var inherits = require('inherits');
+var utils = require('../bemxjst/utils');
+var Entity = require('./entity').Entity;
+var BEMXJST = require('../bemxjst');
+BEMXJST.prototype.Entity = Entity;
```
indutny: Wooot? Why not `BEMHTML.prototype.Entity = Entity`?

lib/bemtree/index.js
```diff
@@ -0,0 +1,55 @@
+var inherits = require('inherits');
+var BEMXJST = require('../bemxjst');
+var Entity = require('./entity').Entity;
+BEMXJST.prototype.Entity = Entity;
```
indutny: Ditto.

lib/bemxjst/entity.js
tadatuta: I still not sure about the whole split of flushing between bemhtml and bemtree (which actually shouldn't use flush at all). Will it be ok to just move `this.canFlush = true;` to bemxjst constructor?

tadatuta:
:up:

lib/bemxjst/entity.js
indutny: You can't do it there, because it is overridden by `.xjstOptions` in user templates. (This is what `this.options` is here for).

tadatuta: like like it was done this way before: https://github.com/bem/bem-xjst/blob/master/lib/bemhtml/runtime/entity.js#L15, wasn't it?

what is the proper way then?

indutny: Oh, you just fixed it. Looks better now, thanks!

indutny:
LGTM, but please squash the commits :)

tadatuta:
decided to move following chages into separate PRs:
* compiler: Drop wrap method
* package.json: Remove bundle command
* travis: Run tests on node 4

PR #149, opened 2015-12-14
BEMHTML: mods field in BEMJSON should not be treated as elemMods

### LINKED ISSUE #88: mods field in BEMJSON should not be treated as elemMods
For BEMJSON `{ block: 'b1', elem: 'e1', mods: { m1: 'v1' } }` with default templates the result is
`<div class="b1__e1 b1__e1_m1_v1"></div>` but it should be `<div class="b1__e1"></div>`.

That'll break backward compatibility (with all previous implementations) so it should be major version.

For discussion please refer to https://github.com/bem/bem-core/issues/1026#issuecomment-128480156

// cc @veged @narqo

tadatuta:
Closes #88

### DIFF
--- a/lib/bemhtml/index.js
+++ b/lib/bemhtml/index.js
@@ -99,7 +99,7 @@ BEMHTML.prototype.render = function render(context,
 
   out += ' class="';
   if (isBEM) {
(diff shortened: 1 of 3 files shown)


indutny:
LGTM

Issue #111, opened 2015-10-28
Should support changing elemMods in runtime


tadatuta:
There is a test for changing `mods` in runtime (https://github.com/bem/bem-xjst/blob/master/test/runtime-test.js#L221) but same thing fails for `elemMods`. So this test 
```js
it('should support changing elemMods in runtime', function() {
      test(function() {
        block('b1').elem('e1').def()(function() {
          this.elemMods.a = 'b';
          return applyNext();
        });
      }, {
        block: 'b1',
        content: { elem: 'e1' }
      }, '<div class="b1"><div class="b1__e1 b1__e1_a_b"</div>');
    });
```
will fail with `AssertionError: '<div class="b1"><div class="b1__e1"></div></div>' deepEqual '<div class="b1"><div class="b1__e1 b1__e1_a_b"</div>'`.

zxqfox:
Take a look pleasee?

/cc @indutny @veged

veged:
looks ok

PR #150, opened 2015-12-16
bemhtml: fix elemMods behaviour

zxqfox:
Fixes: #111

PR #157, opened 2015-12-22
bemhtml: fix elemMods behaviour

miripiruni:
Fixes: #111

P.S. Absolutely identical with https://github.com/bem/bem-xjst/pull/150


zxqfox:
:+1:

miripiruni:
Just remove unnecessary line https://github.com/bem/bem-xjst/pull/150/files#diff-babcc1050af43cb83540193c8c57664dR407

miripiruni:
JFYI The fix depends on 85e02cc and I cannot release it in 4.4.0. Will be released in 5.0.0

Issue #146, opened 2015-12-09
Do not execute nested match if outer one is falsy


tadatuta:
E.g. for
```js
block('bla')(
    tag()('span'),
    match(function() { return this.ctx.d; })( // this.ctx.d is undefined
        tag()('a'),
        attrs()(match(function() { return this.ctx.d.a; })(function(){ // this will throw error
            return { f: 1 };
        }))
    )
) 
```

cc @indutny @veged @belyanskii

veged:
semantically there is no such thing as "nested match" — they are all same level and order of execution is not guaranteed

tadatuta:
Such code worked properly on `1.x` so we should at least try to fix it or maybe support it on convertor side.

veged:
@indutny could we at least support preserving order of `.match`es execution?

PR #158, opened 2015-12-22
bemxjst: fix order of execution of `match`es

indutny:
Always execute matches in the same order as they came in the input file.

Fix: #146

### DIFF
--- a/lib/bemxjst/match.js
+++ b/lib/bemxjst/match.js
@@ -107,11 +107,8 @@ function MatchTemplate(mode, template) {
   }
 
   // Insert late predicates
-  for (var i = postpone.length - 1; i >= 0; i--)
-    this.predicates[i + j] = this.predicates[i];
-  for (var i = 0; i < postpone.length; i++)
-    this.predicates[i] = postpone[i];
-  j += postpone.length;
+  for (var i = 0; i < postpone.length; i++, j++)
+    this.predicates[j] = postpone[i];
 
   if (this.predicates.length !== j)
     this.predicates.length = j;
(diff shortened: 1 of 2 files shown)


lib/bemxjst/match.js
```diff
@@ -186,7 +183,7 @@ Match.prototype.exec = function exec(context) {
   for (var i = 0; i < this.count; i++) {
     if ((mask & bit) === 0) {
       template = this.templates[i];
-      for (var j = template.predicates.length - 1; j >= 0; j--) {
+      for (var j = 0; j < template.predicates.length; j++) {
```
zxqfox: `(var j = 0, l = template.predicates.length; j < l; j++)`

lib/bemxjst/match.js
```diff
@@ -195,7 +192,7 @@ Match.prototype.exec = function exec(context) {
       }
 
       // All predicates matched!
-      if (j === -1)
+      if (j === template.predicates.length)
```
zxqfox: `if (j === l)`

lib/bemxjst/match.js
indutny: Nah, compiler will do this for us! :)

lib/bemxjst/match.js
indutny: V8 should handle this too.

zxqfox: But IE?!

indutny: New IE will do it quite fine, old IE are kind of slow anyway? Do we want to have crappy code just because of one old browser?

indutny: (I sincerely hope that your last comment was sarcasm)

zxqfox: Uh-oh, okay! ;-D

zxqfox:
:+1:

tadatuta:
@indutny tests passed, shell we merge?

indutny:
@tadatuta please do it :+1:

tadatuta:
thank you!

Issue #152, opened 2015-12-16
Should ignore `elem: ''` 


blond:
BEMJSON:

```js
{
    block: 'b1',
    elem: 'e1'
}
```

BEMHTML:

```js
block('b1').elem('e1').def()(function() {
    return applyCtx(this.extend(this.ctx, {
        block: 'b2',
        elem: ''
    }));
});
```

Actual:

```html
<div class="b2__"></div>
```

Expected:

```html
<div class="b2"></div>
```

PR #159, opened 2015-12-22
class-builder: `elem === ''` means no elem

indutny:
Fix: #152

### DIFF
--- a/lib/bemxjst/class-builder.js
+++ b/lib/bemxjst/class-builder.js
@@ -5,7 +5,7 @@ function ClassBuilder(options) {
 exports.ClassBuilder = ClassBuilder;
 
 ClassBuilder.prototype.build = function build(block, elem) {
-  if (elem === undefined)
+  if (!elem)
     return block;
   else
     return block + this.elemDelim + elem;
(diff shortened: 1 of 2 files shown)


zxqfox:
We just though about classBuilder as a https://github.com/bem/bem-naming clone. What you think about using `.stringify` in bem-xjst?

indutny:
I'm fine with this as long as benchmarks are happy with `.stringify`

miripiruni:
IMHO, I would like to merge PR as is. And open another one for [modVal and elemModVal](https://github.com/bem/bem-xjst/commit/6e05992e14340174e23d326831677988d50b7295).

zxqfox:
Yeah, agree.

tadatuta:
lgtm

lib/bemxjst/class-builder.js
```diff
@@ -5,7 +5,7 @@ function ClassBuilder(options) {
 exports.ClassBuilder = ClassBuilder;
 
 ClassBuilder.prototype.build = function build(block, elem) {
-  if (elem === undefined)
+  if (!elem)
```
zxqfox: Guess we can add here `elem !== 0` now or later?

zxqfox:
Lgtm for now

Issue #135, opened 2015-12-02
Reimplement elemMatch


tadatuta:
`elemMatch` should work as follows:

```js
block('b1').elemMatch(function() {
    return this.elem === 'e1' || this.elem === 'e2';
}).tag()('span');
```
should be equal to

```js
block('b1').elem('*').match(function() {
    return this.elem === 'e1' || this.elem === 'e2';
}).tag()('span');
```

miripiruni:
See #184

zxqfox:
Houston: http://goo.gl/U8JpEu — is this ok? I think we need a fix for 4.x

PR #165, opened 2015-12-28
Reimplement elemMatch

miripiruni:
For #135

/cc @zxqfox @tadatuta  @indutny

### DIFF
--- a/test/bemhtml/tree-test.js
+++ b/test/bemhtml/tree-test.js
@@ -175,6 +175,51 @@ describe('BEMHTML compiler/Tree', function() {
     }, '<div class="b1__e1 b1__e1_key_val">ok</div>')
   });
 
+  it('should support elemMatch() match', function() {
+    test(function() {
+      block('b1').elemMatch(function() {
+        return this.elem === 'e1' || this.elem === 'e2';
+      }).tag()('b');
+    }, [
+      { block: 'b1', elem: 'e1' },
+      { block: 'b1', elem: 'e2' },
+      { block: 'b1', elem: 'e3' }
+    ], '<b class="b1__e1"></b><b class="b1__e2"></b><div class="b1__e3"></div>')
+  });
+
+  it('priority of elemMatch and elem is determined by templates order',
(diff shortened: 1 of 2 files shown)


zxqfox:
Here is the problem with back-compat:
```
block('b').def()(
  elemMatch(function(){ return this.elem === 'e1'; })('1'),
  elem('e2')('2')
);
// +
{ block: 'b', elem: 'e2' }
```

miripiruni:
//cc @veged

miripiruni:
@zxqfox Here is two [tests](https://github.com/bem/bem-xjst/pull/165/files#diff-e241c0081ba5875bf6cd84d3eb063aebR190) about priority.

test/bemhtml/tree-test.js
```diff
@@ -175,6 +175,49 @@ describe('BEMHTML compiler/Tree', function() {
...
+        return this.elem === 'e1' || this.elem === 'e2';
+      }).tag()('b');
+    }, [
+      { block: 'b1', elem: 'e1' },
+      { block: 'b1', elem: 'e2' },
+      { block: 'b1', elem: 'e3' }
+    ], '<b class="b1__e1"></b><b class="b1__e2"></b><div class="b1__e3"></div>')
+  });
+
+  it('priority of elemMatch is higher than elem', function() {
```
zxqfox: Actually, I don't get why?

test/bemhtml/tree-test.js
```diff
@@ -175,6 +175,49 @@ describe('BEMHTML compiler/Tree', function() {
...
+    test(function() {
+      block('b1').def()(
+         elem('e1')('elem'),
```
zxqfox: `e1` should be a `e` — that's why. ;-)

arikon:
We decided to remove `elemMatch()` in 5.0 in favor of `elems('a', 'b', 'c', ...)`

zxqfox:
Don't forget to remove or fix `elem('*')`

Issue #172, opened 2016-01-14
Should properly render attrs


miripiruni:
See [test cases](https://github.com/bem/bem-xjst/pull/183/files#diff-958f90d6d94bb3ea244d8d1787058c15).

PR #167, opened 2015-12-30
BH specs

zxqfox:

PR #183, opened 2016-01-21
BEMHTML: should properly render attr values

miripiruni:
Fix for #172

### DIFF
--- a/test/runtime-test.js
+++ b/test/runtime-test.js
@@ -710,6 +710,41 @@ describe('BEMHTML compiler/Runtime', function() {
       test(function() {
       }, { tag: 'span', attrs: { a: 'b' } }, '<span a="b"></span>');
     });
+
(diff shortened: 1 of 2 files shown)


test/runtime-test.js
```diff
@@ -710,6 +710,41 @@ describe('BEMHTML compiler/Runtime', function() {
       test(function() {
       }, { tag: 'span', attrs: { a: 'b' } }, '<span a="b"></span>');
     });
+
+    it('should properly render undefined as attrs value', function () {
```
zxqfox: I'd wrap this all into the describe.

miripiruni: @zxqfox I can do it, but I think it’s useless because of https://github.com/bem/bem-xjst/pull/167

veged:
LGTM

PR #192, opened 2016-02-02
New tests

miripiruni:
v4.x branch: 106 tests. This branch: 291 tests.

 - [x] All old tests was reorganized in [structure](https://gist.github.com/miripiruni/e169331826ce8ed9f17d).
 - [x] More tests added.
 - [x] `mocha` updated to 2.4.5
 - [x] All bugs from this tests have label [From BH tests](https://github.com/bem/bem-xjst/labels/From%20BH%20tests).
 - [x] `npm run lint` now run JSHint on `./test/*.js`

As bonus:
 - [x] You can run tests separately.
 - [x] Error in the fall tests become more understandable (there is a link to a file).


tadatuta:
lgtm

zxqfox:
lgtm too

miripiruni:
May I merge?

indutny:
I have several concerns about this PR:

1. (nit) uppercase letters in path are kind of lame, I don't see any value in having them here, except that it makes tab completion a bit harder on linux (Let's punish those linux guys!)
2. Having `asserts-test.js` is against the test logic. These asserts are regression tests for the functionality that is tested in nearby functions. If you group them together and put in one file, it won't really help anyone, in my opinion.
3. `applyNext`/`applyCtx`/`local`does not really belong to `BEMContext`, I guess? It is more likely a part of runtime

miripiruni:
@indutny thank you.

1. My bad. Fixed.
2. I agree with you. Fixed.
3. Agree. Fixed.

test/bemhtml-test.js
```diff
@@ -0,0 +1,107 @@
+var fixtures = require('./fixtures')('bemhtml');
+var test = fixtures.test;
+
+describe('BEMHTML compiler/Tree', function() {
```
indutny: `Tree`?

indutny:
LGTM

PR #196, opened 2016-02-15
Throw away jsAttr()

miripiruni:
1. `jsAttr()` does not work at all.
2. I don’t know any usecase for using `jsAttr()`.

//сс @tadatuta


tadatuta:
lgtm

zxqfox:
:+1:

PR #200, opened 2016-02-27
Pass BEMContext instance as argument to template body

### LINKED ISSUE #171: ES6-sugar for templates
Is it possible to pass node and/or ctx as parameters to matcher/template to use arrow functions in templates:
```js
block('*').match(node => node.url).tag()('a');
block('*').cls()(node => `position${node.position}`);
block('*').content()({ctx} => [ctx.text, applyNext()]);
```

zxqfox:
Fixes #171 

Was not done:
```js
block('*').content()((_, ctx) => [ctx.text, applyNext()]);
```
because it's a sugar for:
```js
block('*').content()({ctx} => [ctx.text, applyNext()]);
```

I'm not sure what's sweeter.

### DIFF
--- a/lib/bemxjst/match.js
+++ b/lib/bemxjst/match.js
@@ -48,7 +48,7 @@ function MatchCustom(template, pred) {
 }
 
 MatchCustom.prototype.exec = function exec(context) {
-  return this.body.call(context);
+  return this.body.call(context, context, context.ctx);
 };
 
 function MatchOnce(template) {
@@ -167,7 +167,7 @@ Match.prototype.push = function push(template) {
 
 Match.prototype.tryCatch = function tryCatch(fn, ctx) {
   try {
-    return fn.call(ctx);
+    return fn.call(ctx, ctx, ctx.ctx);
   } catch (e) {
     this.thrownError = e;
   }
--- a/test/runtime-match-test.js
+++ b/test/runtime-match-test.js
@@ -1,0 +1,62 @@
+var fixtures = require('./fixtures')('bemhtml');
+var test = fixtures.test;
+
+describe('Runtime Match', function() {
+  it('should call body function in BEMContext context', function() {
+    test(function() {
+      block('b').def()(function() {
+        return this.constructor.name;
+      });
+    }, { block: 'b' }, 'ContextChild');
+  });
+
+  it('should pass BEMContext as the first argument', function() {
+    test(function() {
+      block('b').def()(function(ctx) {
+        return ctx.constructor.name;
+      });
+    }, { block: 'b' }, 'ContextChild');
+  });
+
+  it('should pass BEMContext instance to custom mode', function() {
+    test(function() {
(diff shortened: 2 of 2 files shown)


miripiruni:
LGTM

lib/bemxjst/match.js
```diff
@@ -48,7 +48,7 @@ function MatchCustom(template, pred) {
 }
 
 MatchCustom.prototype.exec = function exec(context) {
-  return this.body.call(context);
+  return this.body.call(context, context);
```
sullenor: Just wondering about won't it be more convenient way of usage?

```javascript
return this.body.call(context, context.ctx, context);
```

for

```javascript
block('*').content()(ctx => ['hello, ' + ctx.content]);
```

zxqfox: With destructuring you could do:
```js
block('*').content()({ctx} => ['hello, ' + ctx.content]);
```

sullenor: Yeah, I know that. Still I have a small hope to avoid code compilation step, since destructuring assignment hasn't shipped to node.js yet.

Anyways both variants are good and I vote for this feature.

sullenor:
:+1:

lib/bemxjst/match.js
zxqfox: Chrome 49 and  already has that. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Browser_compatibility

Let's just wait a little ;-)

tadatuta:
rebase is needed

test/runtime-match-test.js
```diff
@@ -0,0 +1,50 @@
...
+    test(function() {
+      block('b').def()(function(ctx) {
+        return ctx.constructor.name;
+      });
+    }, { block: 'b' }, 'ContextChild');
+  });
+
+  it('should pass BEMContext instance to custom mode', function() {
+    test(function() {
+      block('b').mode('custom')(function (ctx) {
```
miripiruni: without space: `function(ctx)`

test/runtime-match-test.js
```diff
@@ -0,0 +1,50 @@
...
+    test(function() {
+      block('b').def()(function(_, json) {
+        return json.foo;
+      });
+    }, { block: 'b', foo: 'bar' }, 'bar');
+  });
+
+  it('should pass json to custom mode', function() {
+    test(function() {
+      block('b').mode('custom')(function (_, json) {
```
miripiruni: without space: `function (_, json)`

miripiruni:
What about `match`?
```js
block('b').match((_, json) => json._myFlag)(…);
```

miripiruni:
@zxqfox ping

zxqfox:
@miripiruni pong

miripiruni:
LGTM

veged:
LGTM

Issue #201, opened 2016-02-27
BEMHTML: Git rid of slashes for single tags (or make them optional)


tadatuta:

miripiruni:
I think we need an option `xhtml {Boolean}` with default value: false.

PR #206, opened 2016-03-03
BEMHTML: Make XHTML mode optional. #201

tadatuta:


lib/bemhtml/index.js
```diff
@@ -3,8 +3,9 @@ var utils = require('../bemxjst/utils');
 var Entity = require('./entity').Entity;
 var BEMXJST = require('../bemxjst');
 
-function BEMHTML() {
+function BEMHTML(options) {
   BEMXJST.apply(this, arguments);
+  this._shortTagCloser = options.xhtml === true ? '/>' : '>';
```
zxqfox: кажется, что переменная (не свойство) будет дешевле

miripiruni:
@tadatuta I would suggest get rid of `this._shortTagCloser` and merge.

tadatuta:
I thought of `xhtml = false` by default in new major version (as there's no any reason in extra `/`) but you already released 6.0.0 :)

So what'd you suggest instead of `this._shortTagCloser`?

miripiruni:
In order to make this code faster you can leave `this._shortTagCloser`.

Can we change default value of `xhtml` to `true` and release it as `5.2`?

And in 6.1 we can inverse `xhtml`.

tadatuta:
> And in 6.1 we can inverse xhtml.

Not sure if it's ok to make such changes in minor releases (even though 6.0 was not a real major one). Let's postpone inversion till 7.0?

miripiruni:
@tadatuta oh, yes, of course 7.0

tadatuta:
:up:

miripiruni:
:+1:

miripiruni:
May I merge?

tadatuta:
if you asking me — I'm fine with it :)

test/bemhtml-test.js
```diff
@@ -104,4 +105,48 @@ describe('BEMHTML engine tests', function() {
       block: 'b1'
     }, '<div class="b1">ok</div>');
   });
+
+  describe('xhtml option', function() {
+    it('should close short tags in BEMJSON by default', function() {
+      compile('')
+        .apply({ tag: 'br' })
+        .should.equal('<br/>');
+    });
```
veged: should be enough to make just 2 `it`'s about default with BEMJSON and `false` with BEMHTML

PR #216, opened 2016-03-15
Fix for calc position

miripiruni:
#174 Fixed.

I want to cherry-pick the commit to 4.x and 5.x.

### DIFF
--- a/test/bemcontext-position-test.js
+++ b/test/bemcontext-position-test.js
@@ -4,75 +4,89 @@ var test = fixtures.test;
 describe('BEMContext this.position', function() {
   it('should have proper this.position', function() {
     test(function() {
-      block('b1').content()(function() { return this.position; });
+      block('b').content()(function() { return this.position; });
     }, [
-      { block: 'b1' },
-      { block: 'b1' },
-      '',
-      { block: 'b1' },
-      {
-        tag: 'div',
-        content: 'blah'
-      },
-      { block: 'b1' }
-    ], '<div class="b1">1</div>' +
-      '<div class="b1">2</div>' +
-      '<div class="b1">3</div>' +
-      '<div>blah</div>' +
-      '<div class="b1">4</div>');
+      { block: 'b' },
+      { block: 'b' },
+      { block: 'b' }
+    ], '<div class="b">1</div>' +
+      '<div class="b">2</div>' +
+      '<div class="b">3</div>');
   });
 
-  it('should calc position', function() {
+  it('should not count not bem entities', function() {
     test(function() {
-      block('button').elem('inner').def()(function() {
+      block('b').content()(function() { return this.position; });
+    }, [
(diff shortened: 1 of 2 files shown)


zxqfox:
Need to check cases with nested arrays, nulls, and single item.

Something like:
```js
[ 'before', { block: 'b1' }, [ { block: 'b1' } ], 'a' ]
[ null, { block: 'b1' }, null, { block: 'b1' }, null ]
[ { block: 'b1' } ]
```

miripiruni:
@zxqfox 
nested arrays: https://github.com/bem/bem-xjst/blob/issue_174/test/bemcontext-position-test.js#L58
nulls: https://github.com/bem/bem-xjst/blob/issue_174/test/bemcontext-position-test.js#L17
single item: https://github.com/bem/bem-xjst/blob/issue_174/test/bemcontext-position-test.js#L79

Or you talk about more complex test?

test/bemcontext-position-test.js
```diff
@@ -92,4 +105,14 @@ describe('BEMContext this.position', function() {
     { block: 'b', content: { elem: 'e' } },
     '<div class="b"><div class="b__e">1</div></div>');
   });
+
+  it('should calc position for nested blocks', function() {
+    test(function() {
+      block('*').cls()(function() {
```
zxqfox: прямо в класс!?

zxqfox:
Can we add skippage counting for objects without block/elem fields: `{ block: undefined, elem: undefined, text: 'I‘m not a BEM entity' }`?

:ok_hand: at all!

miripiruni:
@tadatuta what do you think? May I merge it?

tadatuta:
I doubt if it's really patch. Potentially it may break someone's project.

miripiruni:
@tadatuta oh, it’s complicated. It’s bug. Shall we treat it as major?

tadatuta:
@miripiruni when it was introduced first?

miripiruni:
Break counting from 2? ;)

zxqfox:
Well, potentially, "someone" who face this problem first should have been reported it. I assume there are no "someone"s since there are no reports from them. Anyway, it's not a behaviour change, it's a fix. It's definitely not a major.

tadatuta:
@zxqfox that someone may think it's just ok. knowledge that we count BEM entities and not any node — is just inside our heads, users don't know it should work this way, they try and see it counts everything.

so there could be templates which use current behaviour as a feature or implement fixes on project level which will fail after this PR.

zxqfox:
No, sorry, but when you have `position = 2` in `{block: 'alone'}` — it's not ok. Definitely.

tadatuta:
well, can't argue with that :)

veged:
LGTM

miripiruni:
Docs needed.

PR #235, opened 2016-04-02
virtual DOM engine

awinogradov:
Original PR https://github.com/awinogradov/bem-xjst/pull/1. Read conversation there.


lib/vidom/index.js
```diff
@@ -0,0 +1,375 @@
...
+}
+
+inherits(VIDOM, BEMXJST);
+module.exports = VIDOM;
+
+VIDOM.prototype.Entity = Entity;
+
+VIDOM.isOneElement = function(children) {
+  return !Array.isArray(children) ||
+      children.length > 1 &&
```
miripiruni: 2 spaces indent

lib/vidom/index.js
```diff
@@ -0,0 +1,375 @@
...
+function VIDOM() {
+  BEMXJST.apply(this, arguments);
+}
+
+inherits(VIDOM, BEMXJST);
+module.exports = VIDOM;
+
+VIDOM.prototype.Entity = Entity;
+
+VIDOM.isOneElement = function(children) {
```
miripiruni: Can you move this function in `VIDOM.prototype.render`?

lib/vidom/index.js
```diff
@@ -0,0 +1,376 @@
...
+        viArguments.push(children);
+      } else {
+        viArguments = viArguments.concat(children);
+      }
+    }
+
+    return viArguments;
+  }
+
+  if (isBEM) {
```
miripiruni: `else if`

lib/vidom/index.js
```diff
@@ -0,0 +1,376 @@
...
+                  mixEntity.block || entity.block,
+                  modName,
+                  mods[modName])
+            );
+          }.bind(this));
+        }.bind(this));
+      }.bind(this));
+    }
+  }
+  if (cls) props.className.push(cls);
```
miripiruni: `else if`

lib/vidom/index.js
```diff
@@ -0,0 +1,376 @@
...
+
+        if (!nestedMix) {
+          return;
+        }
+
+        if (!Array.isArray(nestedMix)) {
+          nestedMix = [ nestedMix ];
+        }
+
+        nestedMix.forEach(function(mix) {
```
miripiruni: for (…) {
  …
}

lib/vidom/index.js
```diff
@@ -0,0 +1,376 @@
...
+          var mods = mix.mods || mix.elemMods;
+          if (!mods) {
+            return;
+          }
+
+          Object.keys(mods).forEach(function(modName) {
```
miripiruni: for

test/vidom/runtime-test.js
```diff
@@ -0,0 +1,1208 @@
...
+
+        block('b1').match(function() {
+          return this.wtf;
+        }).content()('ok');
+      },
+      { block: 'b1' },
+      [ 'div', { className: 'b1' }, 'ok' ]);
+    });
+
+    it('should support > 31 templates (because of the bit mask)', function() {
```
miripiruni: Redundant test.

test/vidom/tree-test.js
```diff
@@ -0,0 +1,388 @@
...
+      block('a')(
+        content()(function() {
+          return [ 'hi', ' there' ];
+        })
+      );
+    },
+    { block: 'a' },
+    [ 'div', { className: 'a' }, 'hi there' ]);
+  });
+});
```
miripiruni: We don’t need to have all of this copy-paste tests.

I think, we need write tests for `lib/vidom/index.js` only.

Guria:
https://github.com/bem-contrib/react-xjst/issues/2
If calculated result is a single 'string', should wrap it with `['div, null, 'string']`

Guria:
```js
  it('should wrap single strings with span', function() {
    test(function() {},
      'hi there',
      [ 'span', null, 'hi there' ]);
  });
// pass

  it('should wrap top-level tag false with span', function() {
    test(function() {
      block('a')(
        tag()(false),
        content()('hi there')
      )
    },
      { block: 'a' },
      [ 'span', null, 'hi there' ]);
  });
// fail
```

awinogradov:
@Guria thanks! We'll fix it)

Guria:
Can't wait a release. Cool stuff.

test/vidom/runtime-test.js
```diff
@@ -0,0 +1,1208 @@
...
+      [ 'div', { className: 'b1__e1' }, 's' ],
+      's2',
+      [ 'div', { className: 'b1__e2' }, 's' ] ]
+    )
+  });
+
+  it('should wrap text into container', function() {
+    test(function() {},
+      [ 'str1', 'str2' ],
+      [ 'div', null, [ 'span', null, 'str1' ], [ 'span', null, 'str2' ] ]);
```
Guria: sholdn't be: `[ 'span', null, 'str1str2' ]`?

Guria:
@awinogradov what about event handlers?
```js
{
    block: 'button',
    name: 'name',
    attrs: { onClick: function(e) {
        console.log(e.nativeEvent)
    } }
}
```

Guria:
make props field to be alias for attrs

awinogradov:
This engine is not about react. We have adapter for react-transformations

awinogradov:
all works fine;) see examples in awinogradov/react-xjst-example

Guria:
Ok :)

lib/vidom/index.js
nik-kor: fixed

test/vidom/tree-test.js
nik-kor: We copy-pasted it from tests for bemhtml engine. And it seems legit because we want to provide the same functionality as bemhtml engine does.

test/vidom/runtime-test.js
nik-kor: I suppose you meant
```
[ 'span', null, 'str1', 'str2'  ]
```

It depends on how we want to make virtual dom later.
It seems like wrapping each string in container is easier to handle later.

nik-kor: but maybe it's a good idea to return [ 'span', null, 'str1', 'str2'  ]
We'll try to fix it

nik-kor:
@Guria event handlers should work

nik-kor:
@Guria fix for bem-contrib/react-xjst#2 3e6a5c778e

miripiruni:
So, what status here?

awinogradov:
Ready, steady, merge!:)

nik-kor:
the only question that I know is https://github.com/bem/bem-xjst/pull/235#discussion_r58512436

lib/vidom/index.js
miripiruni: if (!isBEM && !cls) {
...
} else if (isBEM) {
...
}...

nik-kor: fixed

nik-kor:
@miripiruni I made them fitter https://github.com/bem/bem-xjst/pull/235/commits/dc90b3cc127d1200909c4b40d20e660693af3f97 . 

But, you know, I'm not sure how to check if the particular test is ok. Because these tests are functional and we want to be sure that VIDOM works like BEMHTML and all tests seems useful.

But maybe we should try another tactic and e.g. rewrite all the tests from scratch or make them like unit-tests. 

Resume: it's hard to identify if tests is useful or not.

Guria:
pr currently have at least 2 react specifics: test for `isReactElement` and `className` as `class` attribute.
my suggestions:
* allow to pass `isVirtualNode` function from consumer
* use `class` attribute name by default and rename it to `className` in `react-xjst` package.

Guria:
btw, could we have option to keep `i-bem` classname and js attrs for blocks with js mod?

lib/vidom/index.js
```diff
@@ -0,0 +1,377 @@
...
+      props[name] = utils.attrEscape(
+        utils.isSimple(attr) ? attr : this.context.reapply(attr)
+      );
+    }
+  }
+
+  return props;
+};
+
+VIDOM.prototype.buildClasses = function buildClasses(block, elem, mods) {
```
zxqfox: дублирование же

test/fixtures.js
```diff
@@ -1,6 +1,5 @@
 var bemxjst = require('../');
 var assert = require('assert');
-
```
zxqfox: unnecessary

test/vidom/runtime-test.js
```diff
@@ -0,0 +1,716 @@
...
+describe('VIDOM compiler/Runtime', function() {
+
+  it('should mix className property if passed', function() {
+    test(function() {
+      block('b1')
+    }, { block: 'b1', attrs: { className: 'hi-there' }},
+      [ 'div', { className: 'b1 hi-there' } ]);
+  });
+
+  // FIXME modify this test to check cases with
```
zxqfox: Do we have an issue for that? Could we fix that?

awinogradov:
@Guria I think changing react stuf to i-bem stuf doesn't good idea) But we can return to this talk after merge. Now it's not main problem.

test/vidom/runtime-test.js
awinogradov: Issue for PR? How?)

zxqfox: I'll show you where you can do it easily: https://github.com/bem/bem-xjst/issues/new ;-)

test/vidom/runtime-test.js
```diff
@@ -0,0 +1,716 @@
...
+    test(function() {},
+      { tag: 'section', content: [
+        { block: 'b1' },
+        { block: 'b2' }
+      ] }, [ 'section', null,
+        [ 'div', { className: 'b1' } ],
+        [ 'div', { className: 'b2' } ] ]);
+  });
+
+  // FIXME: waiting for https://github.com/bem/bem-xjst/issues/236
```
zxqfox: Blocker?

test/vidom/runtime-test.js
awinogradov: The issue for unmerged code? Why?

test/vidom/runtime-test.js
awinogradov: It fixed. We need to update tests ;(

test/vidom/runtime-test.js
zxqfox: Because we have a scope of tasks that should be updated. If we gonna merge this we should have an issue about fixing that.

awinogradov: But we can fix it before merge in PR

test/vidom/tree-test.js
```diff
@@ -0,0 +1,372 @@
...
+
+describe('VIDOM compiler/Tree', function() {
+  it('should compile example code', function() {
+    test(function() {
+      block('b1').tag()(
+        elem('e1')('a'),
+        elem('e2').match(function() {
+          return this.ctx.hooray();
+        })(function() {
+          return apply({ a: 1 });
```
zxqfox: It doesn't make sense. It should be something like `.def()` and `apply('default', ...)`. Is it copy-paste? /cc @miripiruni

test/vidom/tree-test.js
```diff
@@ -0,0 +1,372 @@
...
+        })
+      );
+    },
+    { block: 'b1', elem: 'e1' },
+    [ 'a', { className: 'b1__e1' } ]);
```
zxqfox: Tests like this should check for no throw I guess.

Guria:
@awinogradov it is not about changing to `i-bem`. There is no such dom attribute like `className`. I just want to try this engine not only with react, but with some others virtual dom libs. I have thought it was original intention.

awinogradov:
@Guria it's a simple object with props. Not dom attributes only;) I think className is the better name) Because class has more than one description;)

Guria:
@awinogradov bemhtml doesn't have any props as output. only html string which contains of elements which described with tag name and attributes with its values.

awinogradov:
@Guria this PR is not about BEMHTML – it's about Virtual DOM engine for bem-xjst. Try understand BEMTREE engine for example;) It's returns JSON!)))

Guria:
But vidom engine consumes `bemhtml` templates not `bemvidom` one.

awinogradov:
Like bemtree you mean?)

Guria:
`bemhml` templates in `bem-components` could pass props only in js attribute, btw. So I would suggest to split output as separate objects one with dom attributes and second with props.

Guria:
No, I meant that vidom engine doesn't introduces own template dialect `bemvidom`. it was created to reuse `bemhtml` templates to create virtual DOM nodes.

awinogradov:
BEMHTML engine and *.bemhtml doesn't the same things. *.bemhtml it's format of templates which bem-xjst can understand. That's all. Every engine can have diff output because it's target idea. For example: bemhtml – html, bemtree – json, vidom – ddsl. I don't understand problem.

awinogradov:
And of course! Every engine works with *.bemhtml

Guria:
`bemtree` engine doesn't works with `*.bemhtml` templates

awinogradov:
Use http://vladkhvo.github.io

Guria:
it doesn't outputs error, but just not works with all of the `bemhtml` modes:
https://github.com/bem/bem-xjst/blob/master/docs/en/5-templates-syntax.md#bemtree

awinogradov:
All right) Now virtual DOM doesn't support all bemhtml features but we work) I think you understand the idea. One format of templates – many engines. But in the few situations, some engines doesn't need all features of *.bemhtml format. 

Maybe my description is incorrect;( cc @tadatuta 
cc

awinogradov:
Also u shoud know about new bem-xjst. Now it's work by another way.

Guria:
`vidom` engine created to consume `bemhtml` templates to create virtual dom objects instead of `html` strings. So I can't see a room here for properties not tied to DOM or SVG implementation.
It's ok that it's in progress and doesn't support every current `bemhtml`, but you also trying to "extend" it at the same time with some sort of "props". Just let's try to describe output of `vidom` engine.

* tagName
* object with camelCased DOM attributes names as keys
* recursive children nodes declarations

vs 

* tagName
* object with camelCased DOM attributes names as keys, but class replaced with className, because of facebook
* recursive children nodes declarations

------

I hope you see why here is no room for not DOM attributes properties, because `vidom` is substitute for `bemhtml` and not `bemtree`. But if you still want to introduce props in vidom let imagine how it would look like:
```js
// bemjson
{
  block: 'foo',
  mySuperProp: 'whatzaa'
}
// bemhtml
block('foo')(
    attrs()(function () {// have to use attrs mode here to pass it to output
        return {
            mySuperProp: this.ctx.mySuperProp
        }
    })
);
// output
[
    "div",
    {
        "className": "foo",
        "mySuperProp": "whatzaa" // so is it custom attribute or prop?
    }
]
```
http://goo.gl/CJLAiG

awinogradov:
@Guria where is the problem? I really don't understand and why I should force tree twice to solve this problem. Virtual DOM engine returns array of arguments for function wich create the virtual node. Any arguments why they are wrong? Do you know any other functions instead of React where format of properties is different?

awinogradov:
@Guria you should understand. It's not DOM props. It argument of Object type for external function.

awinogradov:
For example) React works with `placeHolder` instead of `placeholder` and etc. Full list of props https://facebook.github.io/react/docs/tags-and-attributes.html#html-attributes

Guria:
At least I want to be it compatible with:
```js
var element = document.createElement(output[0])
Object.keys(output[1]).forEach(function (attr) {
  element.setAttribute(paramCase(attr), output[1][attr])
})
/* ... */
```
because it's about DOM.

> For example) React works with placeHolder instead of placeholder and etc

Wrong: http://www.webpackbin.com/E1Sofmmfb

Again, let's remember your original postulates: 
> Main idea is generation arguments instead of HTML for abstract Virtual DOM API functions. Arguments can be apply in every Virtual DOM implementation tools such as react, vidom, virtual-dom.

> Example arguments:
```js
['div', {...attrs}, content, ...anotherContent]
```
The first argument can be a string with a tag name or existing class of virtual node. The second argument is attributes for the current node. And next arguments are content for the current node. Content can be a string, the same arguments or existing virtual dom node.

https://github.com/awinogradov/bem-xjst/pull/1

> * Написали движок для рендера bemhtml шаблонов и положили аккуратно в PR https://github.com/bem/bem-xjst/pull/235. Экспортит атрибуты в виде массива вида ['div', {...attrs}, content, ...anotherContent].
* Сделали адаптер для React https://github.com/bem-contrib/react-xjst, который использует бандл из шаблонов в bemhtml и инстанс React для создания Virtual DOM , с помощью библиотеки react-dom.

https://ru.bem.info/forum/961/

-----

I am ok if you decide to tie this engine to React specifics and will make clear statement about it. All previous statements was that this engine could be used for every virtual dom lib.

Guria:
> Do you know any other functions instead of React where format of properties is different?

Yeah, lot of them. http://vdom-benchmark.github.io/vdom-benchmark/

>  Any arguments why they are wrong?

All of them was described above. You state that in output willbe attributes, but there is no `className` attribute.

> It's not DOM props. It argument of Object type for external function.

now it's arguments to one specific function implementation.

awinogradov:
Ok, I'm agree. But now we have 2 ways to solve this: do nothing and think about this after merge, wait and support all implementations of virtual DOM. You idea about document.createElement is cool! Can we do it in the future? Not now:( I want do one simple mechanism in this PR. Other things is good but extra features for backlog. Now you can say about wrong name of engine if it support react only. But I think it's good platform for future features. Also of course! If you can solve your wishes by code do it!

Guria:
:+1: I am completely ok with current state, just shared my thoughts about some parts that looks not so consistent. So first comment about react specifics is not so crititical and could be solved later.

What about js attributes. [`bem-components` uses `js` mode](https://github.com/bem/bem-components/search?utf8=%E2%9C%93&q=js+filename%3Abemhtml.js&type=Code) a lot. would be great to deal with this in vidom too. It could also help to mimic lazy initialization as well.

Would be glad to solve it myself, but haven't dived deep into xjst internals yet. If you guys will have no time for this, I'll try it myself, just maybe with your experience you could do it faster.

awinogradov:
@Guria I think lazy loading is not primary task now and in the nearest future. You can help us ;)

lib/vidom/index.js
```diff
@@ -0,0 +1,377 @@
...
+  BEMXJST.prototype.exportApply.call(this, exports);
+};
+
+VIDOM.isAlreadyRendered = function(entity) {
+  return Array.isArray(entity) && entity.length > 1 &&
+    typeof entity[0] === 'string' && typeof entity[1] === 'object' &&
+    entity[1][ALREADY_RENDERED_PROP_NAME];
+};
+
+VIDOM.isReactElement = function(item) {
```
nik-kor: TODO: use ReactElement.isValidElement from react lib

nik-kor:
I think it's better to implement the pluggable system for engines and use it for such engines like VIDOM. Because this PR could take too much time to satisfy the required quality.
Also, VIDOM is interesting for a limited group of people and it's not comparable to BEMHTML now. 

@awinogradov @miripiruni  What do you think?

awinogradov:
@nik-kor I absolutely agree! We talked about this idea with @miripiruni at Hackaton. But now it's impossible to do. We need much time for that. Now @miripiruni do big work to do `bem-xjst` the best of templates engine in the world ;) We can try to do that in the fork and maybe we'll find the right way. Also, I think current PR ready to merge. It is a cool base implementation for React vidom.

nik-kor:
@awinogradov why is it impossible to do? 
it doesn't look so

awinogradov:
@nik-kor it's about time. No tech problems;)

nik-kor:
@miripiruni right. But maybe it will be faster than completing this PR)

Guria:
[mix in bemhtml engine](https://bem.github.io/bem-xjst/?bemhtml=block(%27foo%27).content()(%7B%0A%20%20%20%20elem%3A%20%27bar%27%0A%7D)%0A%0Ablock(%27foo%27).elem(%27bar%27).mix()(function%20()%20%7B%0A%20%20%20%20return%20this.extend(applyNext()%2C%20%7B%20elemMods%3A%20%7B%20bar%3A%20true%20%7D%20%7D)%0A%7D)%0A&bemjson=%7B%0A%20%20%20%20block%20%3A%20%27foo%27%2C%0A%7D)
vs [mix in vidom engine](http://vladkhvo.github.io/bem-xjst/?bemhtml=block(%27foo%27).content()(%7B%0A%20%20%20%20elem%3A%20%27bar%27%0A%7D)%0A%0Ablock(%27foo%27).elem(%27bar%27).mix()(%7B%20elemMods%3A%20%7B%20bar%3A%20true%20%7D%20%7D)%0A%2F%2F%20next%20line%20works%20better%2C%20but%20has%20duplicate%20class%20names%0A%2F%2F%20block(%27foo%27).elem(%27bar%27).mix()(%7B%20block%3A%20%27foo%27%2C%20elem%3A%20%27bar%27%2C%20elemMods%3A%20%7B%20bar%3A%20true%20%7D%20%7D)%0A&bemjson=%7B%0A%20%20%20%20block%20%3A%20%27foo%27%2C%0A%7D%0A)

Guria:
@awinogradov @nik-kor guys, we do not need a fork or this pr merged. we can use `bem-xjst/lib/bemxjst` and do bundling on our side. I'll setup an repo in `bem-contrib`

Guria:
extracted this pr to https://github.com/Guria/xjst-vidom
we can continue to explore there using without affecting this repo
also added a pr to `react-bl` with bundling by webpack

nik-kor:
@Guria great idea)
I like it

Guria:
@awinogradov @nik-kor 
[here is much more simpler engine implementation](http://www.webpackbin.com/EkQMgUIXZ) which can be used with both `React.createElement` and `document.createElement`:
* reused `BEMHTML.prototype.buildModsClasses` and `BEMHTML.prototype.renderMix`.
* dom elements and react elements are created in `def` mode
* `attrs` is converted to `props` with `'react-attr-converter'` module in `def` mode
* dom elements are ready to be inited with `i-bem.js`
* react classes could be also created in `def` mode, though it's a little more complicated and a subject for further explore. see my previous explorations: 
  * https://github.com/awinogradov/react-bl/pull/53/files#diff-d3c31b9613d586be458302467561b1f3R1
  * https://github.com/awinogradov/react-bl/pull/53/files#diff-14e9a41998393a56960970ba15b6e988R78

awinogradov:
@Guria 1 - works here right. But I wanted to use `bem-naming` for all classes building. 2 - why it's a better way? I want extend, not override. 3 - we can implement it here. 4 - useless feature, because we haven't plan to use `i-bem` and `react` at the same time. 

Also (offtopic), I found code generation in runtime in your pr https://github.com/awinogradov/react-bl/pull/54/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR6. I think it's bad way, really. 

@Guria you shouldn't think about me as a bad guy ;) My idea is simple. I want new simple level in BEM terms, where we'll extend base templates with React specific.

And in the end) I think react-bl different task:) Here we want engine wich return `dsDSL` after bemhtml-templates render. `dsDSL` - DSL that is built with the data structures of a programming language.

Guria:
@awinogradov let's count it as workout for my mind. I have learned xjst internals, and it was interesting task to try. So, I am just sharing some my results. I am not trying to insist that my solution is better. But maybe some findings could be still useful.

awinogradov:
@Guria `react-attr-converter` is good. We can use it, but I think `react-xjst` is the better place for normalize.

Guria:
@awinogradov of course, it must not be a part of engine. you can see in my poc engine produces just raw attrs. `react-attr-converter` is compiled to templates on "client-side". I just show that exactly one engine implementation could be used in different appliences without need to have react specifics. Current PR now has lot React specifics and some bugs as well (see [buggy mix reimplemntation](https://github.com/bem/bem-xjst/pull/235#issuecomment-221878143)). I have tried to resolve it by reusing existing methods from `bemhtml` engine.
Again, I am not trying to insist, that my solution is completely right.

Guria:
> Also (off topic), I found code generation in runtime in your PR https://github.com/awinogradov/react-bl/pull/54/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR6. I think it's a bad way, really.

Are you kidding? `webpack.config.js` is not a runtime. How it's different with https://github.com/awinogradov/react-bl/blob/aa1f61832b5e4cdc0f17cb430dc542e71a8bc988/webpack.config.js#L30-L35 ?

awinogradov:
@Guria sorry) yes it's not runtime.

test/vidom/runtime-test.js
Guria: I have moved engine from this pr to own repo, so issues could be pointed there
https://github.com/Guria/xjst-vidom/issues

test/vidom/runtime-test.js
zxqfox: fragments: https://github.com/facebook/react/issues/2127

miripiruni:
@awinogradov can we close it?

awinogradov:
@miripiruni yes, we'll extract tests soon from this PR

awinogradov:
This PR was implemented in separated repo https://github.com/awinogradov/xjst-ddsl with big number if tests https://github.com/awinogradov/xjst-ddsl

PR #237, opened 2016-04-03
Added vidom engine in gh-pages sandbox

vladkhvo:


vladkhvo:
https://github.com/bem/bem-xjst/pull/238 correct ver

Issue #236, opened 2016-04-02
apply('content'), apply('tag') and others must return default values from `this.ctx`


nik-kor:
```js
  it('should be ok', function() {
    test(function() {
      block('b1').def()(function() {
        return applyCtx([apply('content'), this.ctx.content, apply('tag'), this.ctx.tag])
      });
    }, { block: 'b1', tag: 'a', content: '1' },
      'aa11');
  });
```

Проверяли и на версии 6.2

http://bem.github.io/bem-xjst/?bemhtml=%0Ablock(%27b1%27).def()(function()%20%7B%0A%20%20%20%20return%20applyCtx(%5Bapply(%27content%27)%2C%20apply(%27tag%27)%2C%20%27xxx%27%2C%20this.ctx.tag%5D)%0A%7D)%3B&bemjson=%5B%7B%20block%3A%20%27b1%27%2C%20tag%3A%20%271%27%2C%20content%3A%20%27hahaha%27%7D%5D

miripiruni:
Critical

awinogradov:
@miripiruni we can close it, right?

miripiruni:
Fixed in [v6.5.2](https://github.com/bem/bem-xjst/releases/tag/v6.5.2).

PR #255, opened 2016-04-27
`apply(modeName)` must return values from `this.ctx` if no other templates

miripiruni:
Fix for #236 

Benchmarks test results within the error: (+/– 1% in different bundles).

### DIFF
--- a/lib/bemhtml/entity.js
+++ b/lib/bemhtml/entity.js
@@ -15,13 +15,13 @@ function Entity(bemxjst) {
   this.jsClass = null;
 
   // "Fast modes"
-  this.tag = new Match(this);
-  this.attrs = new Match(this);
+  this.tag = new Match(this, 'tag');
+  this.attrs = new Match(this, 'attrs');
   this.mod = new Match(this);
-  this.js = new Match(this);
-  this.mix = new Match(this);
-  this.bem = new Match(this);
-  this.cls = new Match(this);
+  this.js = new Match(this, 'js');
+  this.mix = new Match(this, 'mix');
+  this.bem = new Match(this, 'bem');
+  this.cls = new Match(this, 'cls');
 
   BemxjstEntity.apply(this, arguments);
 }
@@ -50,14 +50,12 @@ Entity.prototype._initRest = function _initRest(key) {
     this.rest[key] = this[key];
   } else {
     if (!this.rest.hasOwnProperty(key))
-      this.rest[key] = new Match(this);
+      this.rest[key] = new Match(this, key);
   }
 };
 
 Entity.prototype.defaultBody = function defaultBody(context) {
   var tag = this.tag.exec(context);
-  if (tag === undefined)
-    tag = context.ctx.tag;
 
   var js;
   if (context.ctx.js !== false)
@@ -68,10 +66,6 @@ Entity.prototype.defaultBody = function defaultBody(context) {
   var mix = this.mix.exec(context);
   var attrs = this.attrs.exec(context);
   var content = this.content.exec(context);
-
-  // Default content
-  if (this.content.count === 0 && content === undefined)
-    content = context.ctx.content;
 
   return this.bemxjst.render(context,
                              this,
--- a/lib/bemhtml/index.js
+++ b/lib/bemhtml/index.js
@@ -69,13 +69,11 @@ BEMHTML.prototype.render = function render(context,
     if (js === true)
       js = {};
 
-    if (js) {
+    if (js && js !== ctx.js) {
       if (ctxJS !== true)
         js = utils.extend(ctxJS, js);
     }  else if (ctxJS === true) {
       js = {};
-    } else {
-      js = ctxJS;
     }
   }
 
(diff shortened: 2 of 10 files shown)


miripiruni:
Updated!

lib/bemhtml/index.js
```diff
@@ -69,13 +69,11 @@ BEMHTML.prototype.render = function render(context,
...
       js = {};
 
-    if (js) {
+    if (js && js !== ctx.js) {
       if (ctxJS !== true)
         js = utils.extend(ctxJS, js);
     }  else if (ctxJS === true) {
       js = {};
-    } else {
-      js = ctxJS;
```
zxqfox: Are you sure about that?

lib/bemhtml/index.js
```diff
@@ -114,16 +112,11 @@ BEMHTML.prototype.render = function render(context,
...
-      else
-        totalMix = ctx.mix;
-    }
+    if (ctx.mix && mix && mix !== ctx.mix)
+      mix = [].concat(mix, ctx.mix);
 
-    if (totalMix) {
-      var m = this.renderMix(entity, totalMix, jsParams, addJSInitClass);
+    if (mix) {
+      var m = this.renderMix(entity, mix, jsParams, addJSInitClass);
```
zxqfox: Looks like redundant change.

bench/package.json
```diff
@@ -5,13 +5,13 @@
...
   "scripts": {
-    "postinstall": "BROWSERIFY=browserify make -C node_modules/bem-xjst",
+    "postinstall": "cd ./node_modules/bem-xjst && npm install && npm run make",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "bem-xjst": "https://github.com/bem/bem-xjst.git#70d77b3",
+    "bem-xjst": "https://github.com/bem/bem-xjst.git#1bd1682",
```
zxqfox: Why not a symlink to `../..`?

miripiruni: @zxqfox symlink? No, we want lock stable (major) revision later.

lib/bemhtml/index.js
miripiruni: That’s OK. `totalMix` is no actual name anymore. there is no any `total` =)

zxqfox: Nice, but how this change related to the original issue and PR?

lib/bemhtml/index.js
miripiruni: Yep. Because of [this.js.exec(context)](https://github.com/bem/bem-xjst/blob/master/lib/bemhtml/entity.js#L64) now returns `this.ctx.js` by default.

zxqfox: 👍

bench/package.json
zxqfox: My bad. Right!

zxqfox:
I'd put totalMix into separate commit, all the rest is LGTM

miripiruni:
@zxqfox separate commit or PR?

zxqfox:
:-D Commit will be enough to me.

lib/bemhtml/index.js
miripiruni: Yes, of course. `this.exec.mix(context)` now returns this.ctx.mix by default. Therefore we don’t need to have all logic in lines 117—123.

miripiruni:
@zxqfox actually, `totalMix` is not needed anymore. See my comment above.

miripiruni:
May I merge?

lib/bemxjst/entity.js
```diff
@@ -18,8 +18,8 @@ function Entity(bemxjst, block, elem, templates) {
   this.canFlush = true;
 
   // "Fast modes"
-  this.def = new Match(this);
-  this.content = new Match(this);
+  this.def = new Match(this, 'def');
```
zxqfox: ?

miripiruni: @zxqfox it looks like not necessary, you’re right.

miripiruni: @zxqfox updated.

miripiruni:
ping

lib/bemhtml/entity.js
```diff
@@ -15,13 +15,13 @@ function Entity(bemxjst) {
...
-  this.tag = new Match(this);
-  this.attrs = new Match(this);
-  this.mod = new Match(this);
-  this.js = new Match(this);
-  this.mix = new Match(this);
-  this.bem = new Match(this);
-  this.cls = new Match(this);
+  this.tag = new Match(this, 'tag');
+  this.attrs = new Match(this, 'attrs');
+  this.mod = new Match(this, 'mod');
```
zxqfox: Didn't noticed before, what the purpose of this? There is no `mod` property in `bemjson` as well as `mod` with template body. We have `block('a').mod('m', 'v')(body)` but not `block('a').mod()(body)`.

zxqfox:
Not sure why what the `mod`. LGTM at all.

lib/bemhtml/entity.js
miripiruni: @zxqfox agree. Thank you. `'mod'` removed.

miripiruni:
OK. 
tumblr_mddkdtuc1a1r9qnhbo1_400

Issue #254, opened 2016-04-27
`mods` disappearing after wrapping an element


zxqfox:
http://goo.gl/Nc01NK

```js
({
  block: 'grid',
  mods: { a: 1 },
  content: { elem: 'cell', elemMods: { a: 1 } }
});
```
+
```js
block('grid').elem('cell').wrap()(function() {
    return {
        block: 'grid',
        elem: 'wrapper',
        content: this.ctx
    };
});

block('grid').elem('cell').content()(function() {
    return 'Expected {"a":1}, but got: ' + JSON.stringify(this.mods)
});
```
→
```html
<div class="grid grid_a_1">
    <div class="grid__wrapper grid__wrapper_a_1">
        <div class="grid__cell">Expected {"a":1}, but got: {}</div>
    </div>
</div>
```

PR #262, opened 2016-05-05
bemjxst: should keep `mods` in case of same block

miripiruni:
Fix for #254

### DIFF
--- a/lib/bemxjst/index.js
+++ b/lib/bemxjst/index.js
@@ -353,7 +353,7 @@ BEMXJST.prototype.runOne = function runOne(json) {
 
     if (json.mods)
       context.mods = json.mods;
-    else
+    else if (json.block !== oldBlock)
       context.mods = {};
   } else {
     if (!json.elem)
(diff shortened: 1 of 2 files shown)


lib/bemxjst/index.js
```diff
@@ -354,7 +354,7 @@ BEMXJST.prototype.runOne = function runOne(json) {
     if (json.mods)
       context.mods = json.mods;
     else
-      context.mods = {};
+      context.mods = json.block === oldBlock ? oldMods : {};
```
zxqfox: `else if (json.block === oldBlock) { context.mods = oldMods } else `?

zxqfox: Ah, no.

just `else if (json.block !== oldBlock) { context.mods = {}; }`

miripiruni: @zxqfox 👍

miripiruni:
🆙

zxqfox:
LGTM

tadatuta:
👍

Issue #190, opened 2016-02-01
BEMHTML: should trim and escape cls


miripiruni:
```js
  it('should trim cls', function() {
    compile(function() {
      block('button').cls()('  btn  ');
    })
      .apply({ block: 'button' })
      .should.equal('<div class="button btn"></div>');
  });

  it('should escape cls', function() {
    compile(function() {
      block('button').cls()('url="a=b&c=d"');
    })
      .apply({ block: 'button' })
      .should.equal('<div class="button url=&quot;a=b&amp;c=d&quot;"></div>');
  });
```

[Demo](http://bem.github.io/bem-xjst/?bemhtml=block(%27button%27).cls()(%27%20%20%20url%3D"a%3Db%26c%3Dd"%20%20%20%27)%3B&bemjson=%7B%20block%3A%20%27button%27%20%7D)

miripiruni:
Released in v6.5.1

PR #264, opened 2016-05-05
BEMHTML: should trim and escape cls

miripiruni:
Fix for #190

### DIFF
--- a/lib/bemhtml/index.js
+++ b/lib/bemhtml/index.js
@@ -130,10 +130,10 @@ BEMHTML.prototype.render = function render(context,
     }
 
     if (cls)
-      out += ' ' + cls;
+      out += ' ' + (cls.trim ? cls.trim() : cls);
   } else {
     if (cls)
-      out += cls;
+      out += cls.trim ? cls.trim() : cls;
   }
 
   if (addJSInitClass)
--- a/lib/bemhtml/index.js
+++ b/lib/bemhtml/index.js
@@ -130,10 +130,11 @@ BEMHTML.prototype.render = function render(context,
     }
 
     if (cls)
-      out += ' ' + (cls.trim ? cls.trim() : cls);
+      out += ' ' + (typeof cls === 'string' ?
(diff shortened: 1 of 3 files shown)


lib/bemhtml/index.js
```diff
@@ -130,10 +130,10 @@ BEMHTML.prototype.render = function render(context,
     }
 
     if (cls)
-      out += ' ' + cls;
+      out += ' ' + (cls.trim ? utils.attrEscape(cls.trim()) : cls);
   } else {
     if (cls)
-      out += cls;
+      out += cls.trim ? utils.attrEscape(cls.trim()) : cls;
```
zxqfox: Ээээээ? What the heck is that condition?

miripiruni: @zxqfox I mean `trim()` exist only in Strings instances.

zxqfox: Em. What else you can pass here?

Why not just to do `cls = utils.attrEscape(cls).trim();` above?

miripiruni: @zxqfox why it’s matter?

zxqfox: It's javascript. `{toString:()=>'">'}` will fall

miripiruni:
@zxqfox updated.

miripiruni:
Review needed.

 cc @zxqfox @tadatuta @veged

lib/bemhtml/index.js
```diff
@@ -130,10 +130,10 @@ BEMHTML.prototype.render = function render(context,
     }
 
     if (cls)
-      out += ' ' + cls;
+      out += ' ' + (cls.trim ? utils.attrEscape(cls).trim() : cls);
```
tadatuta: if you use assert for `cls.trim` just to check if it's an instance of String, why not to do it in more intuitive way: `typeof cls === 'string'`?

tadatuta:
lgtm

PR #267, opened 2016-05-12
BEMHTML: ability to pass attribute style as object (fix #232)

### LINKED ISSUE #232: Ability to pass attribute style as object
Ability to pass object for attribute `style` can be very useful, because we will be able to extend this object and it is cool sugar – we free from necessary concatenate strings.

I think that we should realize next moments:
1. Property name can be specified with minus (`text-align`) or as camel case (`textAlign`)
2. Property value can be specified as number – in this case we mean `px` (except zero)
3. At the end of style list should be a semicolon for avoid sticking in the future

For example:
```js
attrs()({
    style: {
        width: 100, // 100px
        height: 0, // 0
        'text-align': 'center',
        verticalAlign: 'top' // text-align
    }
})
```

And result:
```
style="width:100px;height:0;text-align:center;vertical-align:top;"
```

Example of realization: https://github.com/tenorok/bemer/blob/master/modules/Tag.js#L295-L305
And of tests: https://github.com/tenorok/bemer/blob/master/test/TagTest.js#L136-L153

miripiruni:
For #232 

cc @L0stSoul @tenorok

Just WIP. I need think about it one more time tomorrow.

### DIFF
--- a/lib/bemhtml/index.js
+++ b/lib/bemhtml/index.js
@@ -156,9 +156,16 @@ BEMHTML.prototype.renderClose = function renderClose(prefix,
                                                      content) {
   var out = prefix;
 
-  // NOTE: maybe we need to make an array for quicker serialization
-  attrs = utils.extend(attrs, ctx.attrs);
-  if (attrs) {
+  if (attrs || ctx.attrs) {
+    var ctxStyle = ctx && ctx.attrs && ctx.attrs.style;
+    var style = attrs && attrs.style;
+    style = utils.extend(style, ctxStyle);
+    ctx.attrs = ctx.attrs || {};
+    ctx.attrs.style = style;
+
+    // NOTE: maybe we need to make an array for quicker serialization
+    attrs = utils.extend(attrs, ctx.attrs);
+
     var name; // TODO: do something with OmetaJS and YUI Compressor
     /* jshint forin : false */
     for (name in attrs) {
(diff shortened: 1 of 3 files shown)


tenorok:
Looks fine for me 👍

lib/bemhtml/index.js
```diff
@@ -156,24 +156,45 @@ BEMHTML.prototype.renderClose = function renderClose(prefix,
...
     var name; // TODO: do something with OmetaJS and YUI Compressor
     /* jshint forin : false */
     for (name in attrs) {
       var attr = attrs[name];
       if (attr === undefined || attr === false || attr === null)
         continue;
 
-      if (attr === true)
+      if (name === 'style' &&
+        Object.prototype.toString.call(attr) === '[object Object]') {
```
zxqfox: `attrs: { style: { toString: () => 'omg' } }` isn't supported with this check.

lib/bemhtml/index.js
```diff
@@ -156,24 +156,45 @@ BEMHTML.prototype.renderClose = function renderClose(prefix,
                                                      content) {
   var out = prefix;
 
-  // NOTE: maybe we need to make an array for quicker serialization
-  attrs = utils.extend(attrs, ctx.attrs);
-  if (attrs) {
+  if (attrs || ctx.attrs) {
+    var ctxStyle = ctx && ctx.attrs && ctx.attrs.style;
+    var style = attrs && attrs.style;
```
zxqfox: Guess it's too deep and hacky, isn't it?

lib/bemhtml/index.js
zxqfox: Should we support this in libraries? All will be broken, I guess.

test/modes-attrs-test.js
```diff
@@ -43,4 +43,31 @@ describe('Modes attrs', function() {
...
+        });
+      },
+      { block: 'b' },
+      '<div class="b" style="background:#000;padding:10px 20px;"></div>');
+    });
+
+    it('should merge style from bemjson with templates', function() {
+      test(function() {
+        block('b').attrs()({
+          style: {
```
zxqfox: Why not `block('b').style()({ … });` ?

PR #268, opened 2016-05-13
BEMHTML: Should not render String or Array as attrs value

### LINKED ISSUE #187: BEMHTML: Should not render String or Array as attrs value
```js
  it('should not return String as attrs value', function() {
    test(function() {},
      { block: 'b', attrs: 'skipme' },
      '<div class="b"></div>');
  });

  it('should not return Array as attrs value', function() {
    test(function() {},
      { block: 'b', attrs: [ 1, 2 ] },
      '<div class="b"></div>');
  });
```

miripiruni:
Fix #187 

Benchmarks is OK.

### DIFF
--- a/lib/bemhtml/index.js
+++ b/lib/bemhtml/index.js
@@ -158,10 +158,10 @@ BEMHTML.prototype.renderClose = function renderClose(prefix,
 
   // NOTE: maybe we need to make an array for quicker serialization
   attrs = utils.extend(attrs, ctx.attrs);
-  if (attrs) {
-    var name; // TODO: do something with OmetaJS and YUI Compressor
+  if (attrs && typeof attrs === 'object' && !Array.isArray(attrs) &&
+      attrs !== null) {
     /* jshint forin : false */
-    for (name in attrs) {
+    for (var name in attrs) {
       var attr = attrs[name];
       if (attr === undefined || attr === false || attr === null)
         continue;
--- a/test/bemjson-attrs-test.js
+++ b/test/bemjson-attrs-test.js
@@ -48,8 +48,7 @@ describe('BEMJSON attrs', function() {
       '<div class="b"></div>');
   });
 
-  // TODO: https://github.com/bem/bem-xjst/issues/187
-  xit('should not return String as attrs value', function() {
+  it('should not return String as attrs value', function() {
     test(function() {},
       { block: 'b', attrs: 'skipme' },
       '<div class="b"></div>');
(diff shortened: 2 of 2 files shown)


lib/bemhtml/index.js
```diff
@@ -158,7 +158,7 @@ BEMHTML.prototype.renderClose = function renderClose(prefix,
 
   // NOTE: maybe we need to make an array for quicker serialization
   attrs = utils.extend(attrs, ctx.attrs);
-  if (attrs) {
+  if (attrs && Object.prototype.toString.call(attrs) === '[object Object]') {
```
zxqfox: `toString = Function.prototype.call.bind(Object.prototype.toString)` ?

zxqfox:
👍

lib/bemhtml/index.js
miripiruni: @zxqfox IMHO it’s too much… Why I should do this?

miripiruni:
Review needed.

cc @tadatuta @veged

lib/bemhtml/index.js
zxqfox: Because you already did `Object.prototype.toString.call(attrs) === '[object Object]'` instead of `typeof attrs === 'object'`

miripiruni: @zxqfox OK, I’ll rewrite it in more explicit way.

lib/bemhtml/index.js
```diff
@@ -158,7 +158,8 @@ BEMHTML.prototype.renderClose = function renderClose(prefix,
 
   // NOTE: maybe we need to make an array for quicker serialization
   attrs = utils.extend(attrs, ctx.attrs);
-  if (attrs) {
+  if (attrs && typeof attrs === 'object' && !Array.isArray(attrs) &&
+      attrs !== null) {
     var name; // TODO: do something with OmetaJS and YUI Compressor
```
veged: obsolete TODO ;-)

veged:
LGTM

lib/bemhtml/index.js
miripiruni: @veged TODO removed.

lib/bemhtml/index.js
```diff
@@ -158,8 +158,9 @@ BEMHTML.prototype.renderClose = function renderClose(prefix,
 
   // NOTE: maybe we need to make an array for quicker serialization
   attrs = utils.extend(attrs, ctx.attrs);
-  if (attrs) {
-    var name; // TODO: do something with OmetaJS and YUI Compressor
+  if (attrs && typeof attrs === 'object' && !Array.isArray(attrs) &&
+      attrs !== null) {
+    var name;
     /* jshint forin : false */
     for (name in attrs) {
```
veged: should be `for (var name in attrs) {`

also, as I remember, there was one more similar TODO somewhere

miripiruni: > should be for (var name in attrs) {

Updated.

> also, as I remember, there was one more similar TODO somewhere

I can’t find…

Issue #259, opened 2016-05-04
API .generate(): support arrow functions


miripiruni:
https://github.com/bem/bem-xjst/pull/251

`(() => { return 'lol'; }).toString()`

PR #272, opened 2016-05-18
BEMXJST: generate() should support arrow functions

miripiruni:
Fix for #259 

So, the checks have failed [because of Node v0.1 and Node v0.12](https://travis-ci.org/bem/bem-xjst/builds/131134175). 

WIP: I would like to disable ES6 tests for old Node.js. 

cc @zxqfox

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -15,8 +15,25 @@ Compiler.prototype.generate = function generate(code, options) {
     code = '';
 
   // Yeah, let people pass functions to us!
-  if (typeof code === 'function')
-    code = code.toString().replace(/^function\s*\w*\(\)\s*{|}$/g, '');
+  if (typeof code === 'function') {
+    // Examples:
+    //   function () { … }
+    //   function name() { … }
+    //   function (a, b) { … }
+    //   function name(a, b) { … }
+    var regularFunction = /^function\s*[^{]+{|}$/g;
+
+    // Examples:
+    //   () => { … }
+    //   (a, b) => { … }
+    //   _ => { … }
+    var arrowFunction = /^(_|\(\w|[^=>]+\))\s=>\s{|}$/g;
+
+    code = code.toString();
+    code = code.replace(
+      code.indexOf('function') === 0 ? regularFunction : arrowFunction,
+    '');
+  }
 
   var exportName = options.exportName || 'BEMHTML';
   var engine = options.engine || 'BEMHTML';
--- a/test/modes-elemmod-test.js
+++ b/test/modes-elemmod-test.js
@@ -17,8 +17,9 @@ describe('Modes elemMod(elemModName, elemModVal)', function() {
     },
     {
       block: 'b',
-      content: { elem: 'inner',
-      elemMods: { valid: true }
+      content: {
+        elem: 'inner',
+        elemMods: { valid: true }
       }
     },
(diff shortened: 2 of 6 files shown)


lib/compiler.js
```diff
@@ -15,8 +15,14 @@ Compiler.prototype.generate = function generate(code, options) {
...
   // Yeah, let people pass functions to us!
-  if (typeof code === 'function')
-    code = code.toString().replace(/^function\s*\w*\(\)\s*{|}$/g, '');
+  if (typeof code === 'function') {
+    code = code.toString();
+    code = code.replace(code.indexOf('function') === 0 ?
+      // Regular function
+      /^function\s*\w*\(\)\s*{|}$/g :
+      // Arrow function
+      /^\(\)\s=>\s{|}$/g, '');
```
zxqfox: Please add support for `_ => {}`.

Hm, we don't have variables at all in the `function(){ … }` variant.

Btw, there could be spaces between parens. We could switch to `function\s*[^{]+{…` and `(\w|[^=>]+)\s*=>\s*\{` — these ones are relaxed variants for the first one, they could be much more useful for demo.

miripiruni:
@zxqfox BTW, should I declare that bem-xjst need node >= 4.x? (Not v0.12 or lower)

zxqfox:
@miripiruni It's not required, actually, to support both syntaxes, isn't it?

miripiruni:
@zxqfox yep. In this branch I can turn off node 0.12/0.1 on Travis.

zxqfox:
I don't think it's needed for this issue. So 👎 for dropping 0.10, 0.12.

miripiruni:
@zxqfox I would like to disable ES6 tests for old Node.js.

test/modes-elemmod-test.js
```diff
@@ -17,8 +17,9 @@ describe('Modes elemMod(elemModName, elemModVal)', function() {
     },
     {
       block: 'b',
-      content: { elem: 'inner',
-      elemMods: { valid: true }
+      content: {
+        elem: 'inner',
+        elemMods: { valid: true }
```
zxqfox: Why this change?

miripiruni: @zxqfox because of update JSCS. Now JSCS are report about this.

zxqfox: Ah, my bad.

But can we split JSCS update and support for arrow functions?

zxqfox:
There are two options I see:
- emulate arrow functions behaviour with `var arrowFunction = function() {}; arrowFunction.toString = function(){ return "() => { block('a').tag()('a'); }"; }` — I don't see a big deal with tests for one regex.
- move ES6 code to separate file and skip the whole describe for nodes < 4:
  ```js
  (parseFloat(process.versions.node) < 4 ? describe : describe.skip)('ES6 pieces', function() {
    it('arrow functions', ...);
  });
  ```

miripiruni:
@zxqfox second variant should not work because of SyntaxError with parse `() => {… }` in Node.js v0.12/v0.10.

miripiruni:
Updated!

zxqfox:
The second variant should not require files with ES6 syntax so it will work.

zxqfox:
LGTM <3

Issue #274, opened 2016-05-19
Support `mod('m')` and `elemMod('em')` without second argument.


miripiruni:
From [i-bem documentation](https://en.bem.info/technology/i-bem/v2/i-bem-js-mods/#modifiers):

> A simple modifier is a special case when only the presence or absence of the modifier on the block is important, and its value is insignificant. An example is the modifier describing the ”disabled“ state: disabled. A modifier with an unspecified i-bem.js value is interpreted as boolean and automatically assigned the value true.

`mod('m')` is the same as `mod('m', true)`

zxqfox:
> mod('m') is the same as mod('m', true)

Are you sure?

miripiruni:
> A modifier with an unspecified i-bem.js value is interpreted as boolean and automatically assigned the value true.

veged:
no! you refer to `setMod('m1')`, which is exactly shortcut for `setMod('m1', true)`, BUT `mod('m1')` in terms of "matching" is the same as `mod('m1', '*')`

it's not clear from [documentation](https://en.bem.info/technology/i-bem/v2/i-bem-js-decl/#adding-a-modifier-to-a-block) (sorry, that's my fault)

miripiruni:
Oh my…

zxqfox:
Insane! <3 u all, ty for clarification!

miripiruni:
https://github.com/bem/bem-xjst/releases/tag/v7.3.0

PR #275, opened 2016-05-19
Tree: Support `mod(m)` and `elemMod(em)` without second argument.

miripiruni:
Fix for #274


miripiruni:
See https://github.com/bem/bem-xjst/issues/274#issuecomment-220463920

Issue #202, opened 2016-03-02
applyCtx brokes this.position


zxqfox:
Example: http://goo.gl/fcnGpn

zxqfox:
Polite ping. <3

PR #281, opened 2016-05-23
BEMXJST: Fix for #202 applyCtx brokes this.position

miripiruni:
Fix for #202

### DIFF
--- a/lib/bemxjst/index.js
+++ b/lib/bemxjst/index.js
@@ -52,10 +52,14 @@ BEMXJST.prototype.compile = function compile(code) {
     return self._run(self.context.ctx);
   }
 
+  function applyNewCtx() {
+    return self.run(self.context.ctx);
+  }
+
   function applyCtxWrap(ctx, changes) {
     // Fast case
     if (!changes)
-      return self.local({ ctx: ctx }, applyCtx);
+      return self.local({ ctx: ctx }, applyNewCtx);
 
     return self.local(changes, function() {
       return self.local({ ctx: ctx }, applyCtx);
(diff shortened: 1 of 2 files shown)


test/bemcontext-position-test.js
```diff
@@ -115,4 +115,21 @@ describe('BEMContext this.position', function() {
     { block: 'a1', content: { block: 'a2', content: { block: 'a3' } } },
     '<div class="a1 1"><div class="a2 1"><div class="a3 1"></div></div></div>');
   });
+
+  it('should proper calc position with applyCtx()', function() {
+    test(function() {
+      block('a')(
+        cls()(function() { return this.position; }),
+        match(function() { return !this._flag; })
```
Yeti-or: Why not:

zxqfox:
Agree with @Yeti-or 

The rest LGTM

miripiruni:
Updated.

lib/bemxjst/index.js
```diff
@@ -52,10 +52,14 @@ BEMXJST.prototype.compile = function compile(code) {
...
 
+  function applyNewCtx() {
+    return self.run(self.context.ctx);
+  }
+
   function applyCtxWrap(ctx, changes) {
     // Fast case
     if (!changes)
-      return self.local({ ctx: ctx }, applyCtx);
+      return self.local({ ctx: ctx }, applyNewCtx);
```
zxqfox: Can we add case with changes?

lib/bemxjst/index.js
```diff
@@ -52,10 +52,14 @@ BEMXJST.prototype.compile = function compile(code) {
     return self._run(self.context.ctx);
```
miripiruni: Let’s try to save position here.

miripiruni:
Check position with `wrap` & `replace`.

PR #290, opened 2016-06-14
Add escaped object to simple primitives.

Yeti-or:


miripiruni:
@Yeti-or can you provide background info, please?

lib/bemxjst/utils.js
```diff
@@ -51,6 +51,10 @@ exports.isShortTag = function isShortTag(t) {
 
 exports.isSimple = function isSimple(obj) {
   if (!obj || obj === true) return true;
+  if (!obj.block &&
```
miripiruni: Add `!obj.elem` please

Yeti-or:
Idea of this pr:
we have several *simple* objects, including `String`
but when we want to use `EsapedString` it become *complex* object.
Case for this idea:
We want to check for *simple* values in templates. But we need to keep in mind that escaped strings are not simple.
And when we operate this values in templates (for example strings and escaped strings) in most cases we use them similarly.
I think that it would be easier in templates to use just `isSimple` helper, or to create another helper `isSimpleValue` for example

Yeti-or:
@zxqfox what are you thinking about that?

miripiruni:
@Yeti-or OK. Good.

But Travis failed. And see https://github.com/bem/bem-xjst/pull/290#discussion_r67190873

miripiruni:
LGTM

lib/bemxjst/utils.js
```diff
@@ -52,6 +52,10 @@ exports.isShortTag = function isShortTag(t) {
 exports.isSimple = function isSimple(obj) {
   if (!obj || obj === true) return true;
   if (!obj.block &&
+      !obj.elem &&
+      !obj.tag &&
+      !obj.cls &&
+      !obj.attrs &&
       obj.hasOwnProperty('html') &&
       isSimple(obj.html))
```
zxqfox: Я бы порядок сменил, потому что `obj.html` намного чаще будет возвращать `false`.

Issue #291, opened 2016-06-14
test is controversial


Yeti-or:
https://github.com/bem/bem-xjst/blob/a67900859a928aae53a869cbebca72ba3177b5a3/test/runtime-escaping-test.js#L42
test description and it's content are contradictory

miripiruni:
@Yeti-or you right, I’ll rewrite test description.

Yeti-or:
maybe we need two test case here:
1. ` { block: 'b', html: '<unescaped>', content: 'safe text' }`
2. ` { html: '<unescaped>', content: 'safe text' }`

PR #292, opened 2016-06-16
Tests about escaping improved

miripiruni:
Fix for #291

### DIFF
--- a/test/runtime-escaping-test.js
+++ b/test/runtime-escaping-test.js
@@ -39,10 +39,17 @@ describe('Content escaping', function() {
     '<unescaped>');
   });
 
-  it('should ignore other props if html exists', function() {
+  it('should ignore html field if block/elem/cls/attrs/tag exists', function() {
     test(function() {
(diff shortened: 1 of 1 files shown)


miripiruni:
cc @Yeti-or @zxqfox

test/runtime-escaping-test.js
```diff
@@ -39,12 +39,18 @@ describe('Content escaping', function() {
...
   });
 
-  it('should ignore other props if html exists', function() {
+  it('should ignore html field if block/elem/cls/attrs/tag exists', function() {
     test(function() {
     }, { block: 'b', html: '<unescaped>', content: 'safe text' },
     '<div class="b">safe text</div>');
   });
 
+  it('should render html if only content and html fields exists', function() {
```
zxqfox: I'd say: `it should determine unescaped html field if there are no bem-entity properties`.

miripiruni: @zxqfox OK

miripiruni:
Merge?

zxqfox:
👍

PR #296, opened 2016-06-29
Port useful commits to v4.x

miripiruni:
List of commits (patch level):

1. **BUGFIX**: `apply(modeName)` must return values from `this.ctx` if no other templates (port commit 9d230c5)
2. **BUGFIX**: BEMXJST: Fix for calc position (port commit 1cb9d14)
3. **IMPROVEMENT**: BEMHTML: should trim and escape cls (port commit c5acfb4)

I want ship it to users as v4.3.5.


miripiruni:
@zxqfox @tadatuta Review needed.

zxqfox:
Improvement is a minor, isn't it?

zxqfox:
Ah, nvm. It can be treated as a patch. LGTM

Issue #303, opened 2016-07-04
Bug with mix + elem


miripiruni:
From v6.5.2 there is bug:

```js
  it.only('should work!', function() {
    test(function() {
      block('b').elem('e').replace()(function() {
        var mix = [ {
            block: this.block,
            elem: this.elem,
            mods: this.mods
          } ].concat(this.ctx.mix);

        return {
          block: 'replace',
          mix: mix
        };
      });
    }, {
      block: 'b',
      elem: 'e',
      mix: { block: 'bemjson' }
    }, '<div class="replace b__e bemjson"></div>');
  });
```

```js
      AssertionError: '<div class="replace b__e bemjson__e bemjson__e"></div>' deepEqual '<div class="replace b__e bemjson"></div>'
      + expected - actual

      -<div class="replace b__e bemjson__e bemjson__e"></div>
      +<div class="replace b__e bemjson"></div>
```

Also in v4.3.5.

PR #306, opened 2016-07-06
BEMHTML: Fix for #303 undefined nested mix

miripiruni:
Fix for #303


test/runtime-test.js
```diff
@@ -421,6 +421,35 @@ describe('BEMHTML compiler/Runtime', function() {
       }, '<div class="replace b__e sprite sprite_test_opa bemjson"></div>');
     });
 
+    it.only('should work with undefined nested mix', function() {
```
a-x-: @miripiruni only не забыт?

Issue #312, opened 2016-07-15
BEMHTML: should properly render unescaped html field if `tag:false` present


miripiruni:
BEMJSON:
```js
{
    tag: false,
    html: '<script>console.log("hello html");</script>'
}
```

The result should be:
```
<script>console.log("hello html");</script>
```

Current result:
```js
'[object Object]'
```

zxqfox:
Sad thing ;-(

PR #313, opened 2016-07-15
BEMHTML: should properly render unescaped html field if `tag:false` present

miripiruni:
Fix for #312


zxqfox:
Seems like good fix.

LGTM

PR #347, opened 2016-09-13
TRIVIAL: mod removed as useless

miripiruni:
BEMHTML: `Entity.mod` is useless.

See https://github.com/bem/bem-xjst/commit/6280278b04179f89881f62bf6902b498e58d78b8#diff-bad194964bfd200986dd4bcaf15c2303R20


tadatuta:
is was first introduced at the very beginning of new `bem-xjst` (with drop of `xjst` as a base).
see https://github.com/bem/bem-xjst/commit/a60a13741a102878babfe92ad959962f5a991f86

PR #369, opened 2016-10-06
Update default demo example (fix for #368)

miripiruni:
Current example is awful. I try to change it with new possibilities:
— template body parameters instead of `this` and `this.ctx`
— arrow functions
— show `add*()` mode example
— show `match()` power
— add some comments about what it is

DEMO: https://goo.gl/6TbK7a

@tadatuta @zxqfox what you say?

### DIFF
--- a/default_bemhtml.js
+++ b/default_bemhtml.js
@@ -1,30 +1,23 @@
-block('button')(
-    tag()('button'),
-    attrs()(function() {
-        return {
-            role: 'button',
-            name: this.ctx.name
-        };
-    }),
-    content()(
-        function() {
-            var ctx = this.ctx,
-                content = [ctx.icon];
-            'text' in ctx && content.push({ elem: 'text', content: ctx.text });
-            return content;
-        },
-        match(function() { return typeof this.ctx.content !== 'undefined'; })(function() {
-            return this.ctx.content;
-        })
-    )
-);
+block('link')(
+  // Default tag
+  tag()('span'),
 
-block('icon')(
-    tag()('span'),
-    attrs()(function() {
-        var attrs = {},
-            url = this.ctx.url;
-        if (url) attrs.style = 'background-image:url(' + url + ')';
-        return attrs;
-    })
+  // If url exists in BEMJSON
+  match(function (context, json) { return json.url; })(
+    // Then set a tag
+    tag()('a'),
+    // and set href HTML attribute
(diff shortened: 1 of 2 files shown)


miripiruni:
The only question is: may we use arrow functions in demo? Is it understandably?

zxqfox:
Need to decide what to do with naming: `(context, json)`.
Anything else lgtm.

miripiruni:
@zxqfox what about arrow functions? Is it acceptable?

miripiruni:
@zxqfox about naming: since user can choose any names I want to offer not dwell on this.

tadatuta:
we use `(node, ctx)` on own projects

miripiruni:
@tadatuta why first argument is `node`? It’s `this` — context of template. Why second argument is `ctx` (so ambiguous)? It’s current BEMJSON nodeю

zxqfox:
@miripiruni because ctx is var ctx = this.ctx;

miripiruni:
@zxqfox name `ctx` without context of using is so blurry. While `json` is more accurately. Do you think so?

tadatuta:
I don't like `json` at all. `json` is javascript object notation. It's like calling a variable `integer` or `array`.

`node` is actually `bem-xjst` node as a kind of instance with all the data, helpers and its children. And `ctx` is just current BEMJSON context will set of fields to read from.

miripiruni:
@tadatuta 

> I don't like json at all. json is javascript object notation. It's like calling a variable integer or array.

Do not confuse the type and name. Type `JSON` but name `json`. Type `Array` but name `array`.

> node is actually bem-xjst node as a kind of instance with all the data, helpers and its children. And ctx is just current BEMJSON context will set of fields to read from.

There is no instance of data. It’s one context that was created once at start. That’s why I don’t understand your variant `node`.

tadatuta:
> Do not confuse the type and name. Type `JSON` but name `json`. Type `Array` but name `array`.

I don't confuse them, I just don't like such names ;)

> There is no instance of data. It’s one context that was created once at start. That’s why I don’t understand your variant node.

That's not quite so:
1. `this.block`, `this.elem`, `this.mods` are always about current node.
2. With `applyNext({ blah: 42 })` you'll be able to get `this.blah` only inside subtree of the current node.
3. There are not just contextual data but some helper functions in `this` as well.

And that's why I like `node` ;)

PR #379, opened 2016-11-09
test for nested mix

mishanga:
```js
+ expected - actual

-<div class="root"><div class="link mix a__b a__b_c_d a__b a__b_c_d"></div></div>
+<div class="root"><div class="link mix a__b a__b_c_d"></div>
```


miripiruni:
See https://github.com/bem/bem-xjst/pull/381

PR #386, opened 2016-11-28
Provide arguments for extend() mode

### LINKED ISSUE #385: Provide arguments for extend() mode
Not it's impossible to use `extend()` with arrow functions because there's no arguments provided.
See https://goo.gl/DeCuG5

tadatuta:
Fixes #385

### DIFF
--- a/lib/bemxjst/tree.js
+++ b/lib/bemxjst/tree.js
@@ -119,7 +119,7 @@ ExtendMatch.prototype.wrapBody = function wrapBody(body) {
   return function localAdaptor() {
     var changes = {};
 
-    var obj = body.call(this);
+    var obj = body.call(this, this, this.ctx);
     var keys = Object.keys(obj);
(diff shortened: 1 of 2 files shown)


tadatuta:
If it's ok, I'll port it to 7.x as well.

test/runtime-match-test.js
```diff
@@ -77,4 +77,17 @@ describe('Runtime Match', function() {
       });
     }, { block: 'b', foo: 'This is' }, 'This is ContextChild');
   });
+
+  it('should pass BEMContext instance and json to extend body',
```
miripiruni: Can you move test to [test/modes-extend-test.js](https://github.com/bem/bem-xjst/blob/master/test/modes-extend-test.js)?

tadatuta: sure but aren't it consistent with other tests in this file?

miripiruni:
👍

PR #393, opened 2016-12-06
Online demo: trim BEMJSON input

### LINKED ISSUE #320: Playground bug: ignores bemjson started not from 1th line
Online demo: https://bem.github.io/bem-xjst 

Source code of online demo: https://github.com/bem/bem-xjst/tree/gh-pages

The problem: ignores bemjson started not from 1th line

|  |  |
| --- | --- |
|  |  |

Possible solution: trim BEMJSON before `BEMHTML.apply(BEMJSON)`.

mikhailrojo:
Fixes #320
Playground bug: ignores bemjson started not from 1th line #320

### DIFF
--- a/common.blocks/demo/demo.js
+++ b/common.blocks/demo/demo.js
@@ -112,7 +112,7 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce' ], functi
 
     function safeEval(str) {
         try {
-            return (new Function('return ' + str))();
+            return (new Function('return ' + str.split('\n').join('')))();
         } catch(e) {
             return e;
         }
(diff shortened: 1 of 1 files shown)


common.blocks/demo/demo.js
```diff
@@ -112,7 +112,7 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce' ], functi
 
     function safeEval(str) {
         try {
-            return (new Function('return ' + str))();
+            return (new Function('return ' + str.split('\n').join('')))();
```
miripiruni: Just use `trim()` on str.

miripiruni: Just use `trim()` on `str`.

PR #400, opened 2017-01-09
BEMTREE: special unescaped html field from BEMJSON should render as is

### LINKED ISSUE #399: Wrong html field handling in BEMTREE

There's no way to keep unescaped markup when using `BEMHTML.apply(BEMTREE.apply())`.

### Expected Behavior
```js
BEMTREE.apply({ block: 'b1', content: { html: '<br>' } });
// { block: 'b1', content: { html: '<br>' } }
```
### Actual Behavior
```js
BEMTREE.apply({ block: 'b1', content: { html: '<br>' } });
// { block: 'b1', content: '<br>' }
```
### Possible Solution

This [piece of code](https://github.com/bem/bem-xjst/blob/master/lib/bemxjst/index.js#L280-L288) should only be applied to BEMHTML and not to BEMTREE.

### Your Environment

bem-xjst 7.x

miripiruni:
Fixes #399

### DIFF
--- a/lib/bemtree/index.js
+++ b/lib/bemtree/index.js
@@ -54,3 +54,8 @@ BEMTREE.prototype._run = function _run(context) {
   if (!context || context === true) return context;
   return BEMXJST.prototype._run.call(this, context);
 };
+
+BEMTREE.prototype.runUnescaped = function runUnescaped(context) {
+  this.context._listLength--;
+  return context;
+};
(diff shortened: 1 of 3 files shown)


lib/bemtree/index.js
```diff
@@ -54,3 +54,8 @@ BEMTREE.prototype._run = function _run(context) {
   if (!context || context === true) return context;
   return BEMXJST.prototype._run.call(this, context);
 };
+
+BEMTREE.prototype.runUnescaped = function runUnescaped(context) {
```
miripiruni: The method overrides base BEMXJST method.

PR #405, opened 2017-01-18
More tests for coverage

miripiruni:


lib/bemhtml/index.js
```diff
@@ -81,10 +81,7 @@ BEMHTML.prototype.render = function render(context,
 
   var isBEM = bem;
   if (isBEM === undefined) {
-    if (ctx.bem === undefined)
-      isBEM = entity.block || entity.elem;
-    else
-      isBEM = ctx.bem;
+    isBEM = entity.block || entity.elem;
```
miripiruni: Because of `bem` in line 82 is result of `bem()` mode execution. And `bem()` returns `this.ctx.bem` by default.

lib/bemhtml/index.js
```diff
@@ -93,7 +90,7 @@ BEMHTML.prototype.render = function render(context,
 
   var addJSInitClass = jsParams && (
     this._elemJsInstances ?
-      (entity.block || entity.elem) :
+      entity.block :
```
miripiruni: Because of `entity.block` is always available.

lib/bemhtml/index.js
```diff
@@ -284,9 +281,6 @@ BEMHTML.prototype.renderMix = function renderMix(entity,
     context.elem = oldElem;
     context.block = oldBlock;
 
-    if (!nestedMix)
-      continue;
-
```
miripiruni: Because of check in line 278 already the same.

lib/bemxjst/index.js
```diff
@@ -8,7 +8,7 @@ var ClassBuilder = require('./class-builder').ClassBuilder;
 var utils = require('./utils');
 
 function BEMXJST(options) {
-  this.options = options || {};
+  this.options = options;
```
miripiruni: Because of [`options` is always object](https://github.com/bem/bem-xjst/blob/master/lib/compiler.js#L12).

lib/bemxjst/index.js
```diff
@@ -117,10 +117,6 @@ BEMXJST.prototype.compile = function compile(code) {
 
 BEMXJST.prototype.recompileInput = function recompileInput(code) {
   var args = BEMXJST.prototype.locals;
-  // Reuse function if it already has right arguments
-  if (typeof code === 'function' && code.length === args.length)
-    return code;
-
```
miripiruni: Because this case is fantastic:

```
bemxjst.compile(function(block, elem, mod, addMods, content, …) { … })
```

Nobody wants to enumerate all modes.

zxqfox: Not sure. We can generate code with all modes ;-)

zxqfox: nvm, you're right

lib/bemhtml/index.js
tadatuta: sure it's useless case but it works now:
```js
BEMHTML.apply({ elem: 'e1' }); // '<div class="__e1"></div>'
```

test/api-compile-test.js
```diff
@@ -146,5 +146,60 @@ describe('API compile', function() {
         template.apply({ block: 'b1' });
       });
     });
+
+    it('should throw bem-xjst error on template with no block', function() {
```
tadatuta: without block

PR #409, opened 2017-01-29
compile method of template instance should return itself

### LINKED ISSUE #408: compile method of template instance should return itself
```js
require('bem-xjst').bemhtml.compile('')
/*
{ apply: [Function: apply],
  compile: [Function: compile],
  BEMContext: { [Function: ContextChild] super_: [Function: Context] } }
*/
require('bem-xjst').bemhtml.compile('').compile('') // undefined
```

### Expected Behavior

`compile()` method of template instance should return itself

### Actual Behavior

returns `undefined`

tadatuta:
Fixes #408


miripiruni:
See https://github.com/bem/bem-xjst/pull/410

Issue #418, opened 2017-02-22
Match must have an argument


miripiruni:
### Input code or something about issue background

https://goo.gl/kmH8K0

```
{ block: 'b' }
```
```
block('b').match().content()(2);
```

### Expected Behavior

`Error: .match() must have argument`

### Actual Behavior

`<div class="b">2</div>`

//сс @yalovek

PR #419, opened 2017-02-22
TREE: Match must have an argument 

miripiruni:
#418 fixed

### DIFF
--- a/lib/bemxjst/tree.js
+++ b/lib/bemxjst/tree.js
@@ -353,6 +353,10 @@ Tree.prototype.body = function() {
 
 Tree.prototype.match = function() {
   var children = new Array(arguments.length);
+
+  if (!arguments.length)
+    throw new Error('.match() must have argument');
+
   for (var i = 0; i < arguments.length; i++) {
     var arg = arguments[i];
     if (typeof arg === 'function')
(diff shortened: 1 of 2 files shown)


miripiruni:
cc @yalovek

lib/bemxjst/tree.js
```diff
@@ -353,6 +353,10 @@ Tree.prototype.body = function() {
 
 Tree.prototype.match = function() {
   var children = new Array(arguments.length);
+
+  if (!arguments.length)
+    throw new Error('.match() must have argument');
```
zxqfox: `if (!something) throw new Error('error')` is a pure assert pattern: `assert(something, 'error')`
prob we can get some bytes on it.

miripiruni: after gzip:

term 1: `throw new Error('`
term 2: `')`

In your case:

term 1: `assert(something, '`
term 2: `')`

Same shit…

zxqfox: For the point point of view: you dont need test case for `assert(arguments.length)`

miripiruni: @zxqfox m?

zxqfox: nvm, anyway you writing tests for exceptions

tadatuta: it should be **an** argument but I'm not sure if it's significant in this case

PR #425, opened 2017-03-13
TREE: fixed append/prepend content with function body

### LINKED ISSUE #422: appendContent/prependContent escapes html
bemjson
```js
{
    block: 'foo'
}
```

bemhtml
```js
block('foo')(
    prependContent()(() => ({
        elem: 'x', 
        content: 'before'
    })),
    appendContent()(() => ({
        elem: 'x', 
        content: 'after'
    }))
);
```

result
```html
<div class="foo">&lt;div class="foo__x"&gt;before&lt;/div&gt;&lt;div class="foo__x"&gt;after&lt;/div&gt;</div>
```

Live example https://goo.gl/5wXovY

---------

guess appendContent/prependContent is just sugar for `content` + `applyNext`, but this behaviour is unexpected

miripiruni:
Fixes #422 

cc @golyshevd

### DIFF
--- a/lib/bemxjst/tree.js
+++ b/lib/bemxjst/tree.js
@@ -139,9 +139,7 @@ AddMatch.prototype.wrapBody = function(body) {
 };
 
 AddMatch.prototype.appendContentWrapBody = function(body) {
-  var refs = this.refs;
-  var applyCtx = refs.applyCtx;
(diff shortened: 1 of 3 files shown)


miripiruni:
@golyshevd In which version is it necessary? v7.x or v8.x?

PR #428, opened 2017-03-16
BEMXJST: Revert exportApply optimization

tadatuta:
`oninit` should have global context at least for backward compatibility.


miripiruni:
Can you add test for it?

tadatuta:
@dustyo-O found the answer to the magic. it's because of context.
the diff in the PR is equivalent to 
```js
self.oninit[i].bind(null)(exports, { BEMContext: exports.BEMContext });
```

tadatuta:
Added test

PR #432, opened 2017-03-28
Fixed this.restoreDepth call possibly

SlNPacifist:
This possibly fixes memory leaks in Match instances in #433


miripiruni:
v7.7.4
v8.6.10

PR #437, opened 2017-04-17
BEMXJST: should exec new oninit functions on every compile (fix #356)

### LINKED ISSUE #356: `oninit()` must be executed on every `compile`
oninit-stand.js:

``` js
var b = require('bem-xjst'), // bem-xjst v4.x
    template = b.compile(function() {
      oninit(function() { console.log('1111') ; });
      block('b').tag()('span')
    });

template.compile(function() {
  oninit(function() { console.log('2222') ; });
  block('b').mix()('mixed')
});

var html = template.apply({ block: 'b' });

console.log(html);
```

``` bash
$ node oninit-stand.js
```

Actual result:

``` bash
1111
<span class="b mixed"></span>
```

Expected result:

``` bash
1111
2222
<span class="b mixed"></span>
```

Original discussion: https://nda.ya.ru/3RzKeD

cc @veged @tadatuta @zxqfox

miripiruni:
Fixes #356

### DIFF
--- a/lib/bemxjst/index.js
+++ b/lib/bemxjst/index.js
@@ -97,6 +97,13 @@ BEMXJST.prototype.compile = function(code) {
   // Concatenate templates with existing ones
   // TODO(indutny): it should be possible to incrementally add templates
   if (this.tree) {
+    self.BEMContext = self.contextConstructor;
+    for (var i = 0; i < out.oninit.length; i++) {
+      // NOTE: oninit has global context instead of BEMXJST
+      var oninit = out.oninit[i];
+      oninit(self, { BEMContext: self.BEMContext });
+    }
+
     out = {
       templates: out.templates.concat(this.tree.templates),
       oninit: this.tree.oninit.concat(out.oninit)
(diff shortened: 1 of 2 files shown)


lib/bemxjst/index.js
```diff
@@ -97,6 +97,13 @@ BEMXJST.prototype.compile = function(code) {
   // Concatenate templates with existing ones
   // TODO(indutny): it should be possible to incrementally add templates
   if (this.tree) {
+    self.BEMContext = self.contextConstructor;
+    for (var i = 0; i < out.oninit.length; i++) {
+      // NOTE: oninit has global context instead of BEMXJST
+      var oninit = out.oninit[i];
+      oninit(self, { BEMContext: self.BEMContext });
```
zxqfox: Sorry but looks like a bug. Can you add a test with incrementing (or better concatenating) global variable to check quantity (or better order) of `oninit` runs?

miripiruni: I'll try to extract method…

miripiruni:
@zxqfox first of all I check bundle size with current changes: https://travis-ci.org/bem/bem-xjst/jobs/222942140 — 9580 B

After that I extracted method runOninit() and check bundle size again — 9597 B.

miripiruni:
updated

miripiruni:
@zxqfox can you review it please?

zxqfox:
If the size growing up guess we shouldn't do that.
Looks good at all.

PR #442, opened 2017-04-24
BEMHTML: fix render i-bem when mix elem with js (fix #441)

### LINKED ISSUE #441: Add `i-bem` class when mixed element has `js: true` with `elemJsInstances: true`

```js
{
    block: 'b',
    content: {
        elem: 'e1',
        mix: { elem: 'e2', js: true }
    }
}
```

### Expected Behavior

```html
<div class="b"><div class="b__e1 b__e2 i-bem" data-bem='{"b__e2":{}}'></div></div>
```

### Actual Behavior

```html
<div class="b"><div class="b__e1 b__e2" data-bem='{"b__e2":{}}'></div></div>
```
(no `i-bem` class).

bem-xjst version and all relevant details
`bem-xjst 7.x / 8.x`, with `elemJsInstances: true`

miripiruni:
Fixes #441

### DIFF
--- a/lib/bemhtml/index.js
+++ b/lib/bemhtml/index.js
@@ -234,7 +234,9 @@ BEMHTML.prototype.renderMix = function(entity, mix, jsParams, addJSInitClass) {
       js[classBuilder.build(block, item.elem)] =
           item.js === true ? {} : item.js;
       if (!addInit)
-        addInit = block && !item.elem;
+        addInit = this._elemJsInstances ?
+          (item.elem || block) :
+          (block && !item.elem);
     }
 
     // Process nested mixes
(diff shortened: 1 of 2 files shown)


miripiruni:
@tadatuta @zxqfox is there valide case where we want mix mods/elemMods with js?

```js
// mods
{
    block: 'b',
    mix: { mods: {test: 'ok', js: true } }
}

// elemMods

{
    block: 'b',
    elem: 'e',
    mix: { elemMods: {test: 'ok', js: true } }
}
```

hm?

tadatuta:
@miripiruni 
the example you provided is not about mods mixed with js params, it means just boolean js modifier: `_js`.

as for
```js
{
    block: 'b',
    mix: { mods: { test: 'ok' }, js: true }
}
```
it should be valid but quite strange case (there's no reason not to use `js` outside of the mix).

miripiruni:
@tadatuta so, we cover all cases and I can merge it?

tadatuta:
yes, I think so

PR #443, opened 2017-04-24
BEMXJST: run() should revert depth

### LINKED ISSUE #439: Strange bug with array in attrs
https://goo.gl/Tz3oeD

v6.x
v7.x
v8.x

miripiruni:
Fixes #439

### DIFF
--- a/lib/bemxjst/index.js
+++ b/lib/bemxjst/index.js
@@ -290,6 +290,7 @@ BEMXJST.prototype._run = function(context) {
 BEMXJST.prototype.run = function(json) {
   var match = this.match;
   var context = this.context;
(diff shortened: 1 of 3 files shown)


lib/bemxjst/index.js
```diff
@@ -302,6 +303,7 @@ BEMXJST.prototype.run = function(json) {
 
   this.match = match;
   this.context = context;
+  this.depth = depth;
```
miripiruni: The problem was here.

miripiruni: @tadatuta but in previous commit I decided to use `_run()`. Maybe now it’s much better solution…

PR #455, opened 2017-05-17
BEM-XJST: block(*) should not apply to elems

### LINKED ISSUE #256: Should block('*') apply to elems?
Examples: http://goo.gl/SBJszu

miripiruni:
Fixes #256


miripiruni:
P.S. Check order of templates with '*' applying.

miripiruni:
PR is locked and waiting next major release.

PR #467, opened 2017-06-25
feat (Demo): engine and version into URL

### LINKED ISSUE #389: Online demo: get parameters for version
In online demo we need to be able save in URL version and engine of current demo state.

vitkarpov:
Fixes #389 

- drops load the default `xbrowser.bemhtml.js` bundle during the page load: now it loads dynamically corresponding to value from engine select (the last version is default, so nothing's changed here)
- `version` and `engine` into URL, saved on every select change

### Problems

I've encountered a problem. You can face it if change version below 8.4.0, pick BEMTREE and reload the page: `this._engine is not define` instead of rendered bemjson from the bemtree template.

**Why this happens**

The bundle actually doesn't have `BEMTREE` as a global object, so when you change the version within select in runtime everything is okay, because you actually use `BEMTREE` **from the old bunlde**. In current production the first bundle (version 8.5.2) loads every time, so you can't see the effect.

I'm not sure how to deal with it, let's discuss it :)

### DIFF
--- a/common.blocks/demo/demo.js
+++ b/common.blocks/demo/demo.js
@@ -73,10 +73,7 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce', 'queryst
                 store.get('playground') ||
                 this._getState();
 
-            this._templates.setValue(data.template);
-            this._bemjson.setValue(data.bemjson);
-            this._versionSelect.setValue(data.version);
-            this._engineSelect.setValue(data.engine);
+            this._setState(data);
         },
         _getState: function() {
             return {
@@ -85,6 +82,12 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce', 'queryst
                 template: this._getTemplate(),
                 bemjson: this._getBEMJSON()
             };
+        },
+        _setState: function(data) {
+            this._templates.setValue(data.template);
+            this._bemjson.setValue(data.bemjson);
+            this._versionSelect.setValue(data.version);
+            this._engineSelect.setValue(data.engine);
         }
     }, {}));
 
--- a/common.blocks/demo/demo.js
+++ b/common.blocks/demo/demo.js
@@ -1,4 +1,4 @@
-modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce' ], function(provide, BEMDOM, pretty, debounce) {
+modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce', 'querystring' ], function(provide, BEMDOM, pretty, debounce, qs) {
 
     provide(BEMDOM.decl('demo', {
         onSetMod: {
@@ -65,36 +65,20 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce' ], functi
             this._result.setValue(finalCode);
         },
         save: function() {
-            var template = this._getTemplate(),
-                bemjson = this._getBEMJSON(),
-                version = this._versionSelect.getValue(),
-                engine = this._engineSelect.getValue();
-
-            store.set('playground', {
-                version: version,
-                template: template,
-                bemjson: bemjson,
-                engine: engine
-            });
-
-            history.pushState({}, document.title, location.pathname + '?' + [
-                'template=' + encodeURIComponent(template),
-                'bemjson=' + encodeURIComponent(bemjson),
-                'version=' + encodeURIComponent(version),
-                'engine=' + encodeURIComponent(engine)
-            ].join('&'));
+            store.set('playground', this._getState());
+            history.pushState({}, document.title, location.pathname + '?' + qs.stringify(this._getState()));
         },
         _load: function() {
             var data = parseParams(location.search.split('?')[1]) ||
                 store.get('playground') ||
-                this._getDefaultState();
+                this._getState();
 
             this._templates.setValue(data.template);
             this._bemjson.setValue(data.bemjson);
             this._versionSelect.setValue(data.version);
             this._engineSelect.setValue(data.engine);
         },
-        _getDefaultState: function() {
+        _getState: function() {
             return {
                 template: this._getTemplate(),
                 bemjson: this._getBEMJSON(),
--- a/common.blocks/demo/demo.js
+++ b/common.blocks/demo/demo.js
@@ -4,30 +4,29 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce' ], functi
         onSetMod: {
             js: {
                 inited: function() {
-
+                    
+                    var bPage = this.findBlockOutside('page');
+                    
+                    this._versionSelect = bPage.findBlockInside('version-selector');
+                    this._engineSelect = bPage.findBlockInside('engine-selector');
                     this._templates = this.findBlockOn('templates', 'editor');
                     this._bemjson = this.findBlockOn('bemjson', 'editor');
                     this._result = this.findBlockOn('result', 'editor');
-                    this._engine = BEMHTML;
                     this._debouncedOnChange = debounce(this._onChange, 150, this);
 
                     this._templates.on('change', this._debouncedOnChange);
                     this._bemjson.on('change', this._debouncedOnChange);
 
-                    this._load() || this._render();
-
-                    this.delMod('state');
-
-                    setTimeout(function() {
+                    this._versionSelect.on('ready', function() {
+                        this._load();
                         this.setMod('state', 'loaded');
-                    }.bind(this), 150);
-
+                    }, this);
                 }
             }
         },
         _onChange: function() {
             this._render();
-            this._save();
+            this.save();
         },
         _getTemplate: function() {
             return this._templates.getValue();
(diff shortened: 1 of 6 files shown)


common.blocks/demo/demo.js
```diff
@@ -65,37 +64,43 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce' ], functi
...
+                bemjson: bemjson,
+                engine: engine
             });
 
             history.pushState({}, document.title, location.pathname + '?' + [
                 'template=' + encodeURIComponent(template),
-                'bemjson=' + encodeURIComponent(bemjson)
+                'bemjson=' + encodeURIComponent(bemjson),
+                'version=' + encodeURIComponent(version),
+                'engine=' + encodeURIComponent(engine)
```
miripiruni: Can we set `template=…` and `bemjson=…` as last parameters? If the URL would be so long we don’t saw version and engine. And more over if google URL shortcut strips URL we didn't catch that.

common.blocks/demo/demo.js
```diff
@@ -65,37 +64,43 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce' ], functi
...
-
-            return !!(data.template || data.bemjson)
+            var data = parseParams(location.search.split('?')[1]) ||
+                store.get('playground') ||
+                this._getDefaultState();
+
+            this._templates.setValue(data.template);
+            this._bemjson.setValue(data.bemjson);
+            this._versionSelect.setValue(data.version);
+            this._engineSelect.setValue(data.engine);
```
miripiruni: Let’s extract four lines above as `setState()`?

common.blocks/demo/demo.js
```diff
@@ -65,37 +64,43 @@ modules.define('demo', [ 'i-bem__dom', 'pretty', 'functions__debounce' ], functi
...
+                version = this._versionSelect.getValue(),
+                engine = this._engineSelect.getValue();
 
             store.set('playground', {
-                version: this.params.version,
+                version: version,
                 template: template,
-                bemjson: bemjson
+                bemjson: bemjson,
+                engine: engine
```
miripiruni: Can we create `_getState()`?

common.blocks/demo/demo.js
vitkarpov: > If the URL would be so long we don’t saw version and engine.

Do you think somebody can construct URL by her own? I thought that nobody's looking at urls at all...

> And more over if google URL shortcut strips URL we didn't catch that.

Yeah, it's the problem and I've already seen the issue about it, so should be solved apart.

common.blocks/demo/demo.js
vitkarpov: Smart decision, sir! :) Just didn't want to perform it in sake of diff sanity, but now it surely has to be done.

vitkarpov:
@miripiruni btw, pay attention on the value of `version` parameter, it's not actual version but github hash of that version. It's how version select works now, options have hashes as their values and actual versions is just innerText.

Do you think it's okay? imho, it doesn't matter because of nobody's going to look at urls, seems, everybody use `goo.gl`, don't they?

vitkarpov:
@miripiruni /fixed

miripiruni:
> Do you think it's okay? imho, it doesn't matter because of nobody's going to look at urls, seems, everybody use goo.gl, don't they?

I think more usable would be `version=8.6.0` in URL. But we need catch 404. (example: version=0.0.10000). As profit we get:

1) verbose and understandable URL `version` parameter
2) anyone can change it via change URL
3) it’s much shorter than hash commit.


But if you think its complicated please say it.

miripiruni:
I can suggest a hash map, something like this:

```js
{
  '8': 'dfb897dfb9dfb09df8bdf0b98',
  '7': 'fb89dfb9dwehk3a9pas83lds4',
  '8.6.5': 'dfb897dfb9dfb09df8bdf0b98',
  … // etc
}
```

vitkarpov:
It's more complicated, I'd say, but at the same time it's more convenient, for sure. In order to do this I have to refactor version selector.

Try to implement then within this pr, right?

miripiruni:
Yes, please

vitkarpov:
@miripiruni https://github.com/bem/bem-xjst/pull/467/commits/43fd62f9bb1b61d203484c8bf97f2f0a44c4f33e

- version by name, not hash
- move version and engine to the beginning of the url, not end (as you mentioned earlier)

About https://github.com/bem/bem-xjst/pull/467#issuecomment-310910145 , I'm not sure how to do this, I guess, `update` script should be changed?

vitkarpov:
@miripiruni in this pr I broke a thing: invalidation localstorage cache. It's not very convenient anyways, as I get it right, you have to change postfix for `BEMXJST_VERSION` right into bemjson of the page.

I think it's better to have a param in URL to invalidate cache, what do you think?

vitkarpov:
@miripiruni ping? :)

common.blocks/demo/demo.js
miripiruni: @vitkarpov so, `_setState()` would be nice?

common.blocks/version-selector/version-selector.js
```diff
@@ -20,15 +20,15 @@ modules.define('version-selector', [ 'i-bem__dom', 'querystring' ], function(pro
...
-                            ver = ver['version-selector'].default.hash;
+                            ver = select.params.default.name;
                         }
 
                         select.setValue(ver);
                     };
 
                     select.domElem.on('change', function(e) {
-                        var val = e.target.value;
+                        var name = e.target.value;
```
miripiruni: Why `name`? It’s `version`, right?

common.blocks/version-selector/version-selector.js
```diff
@@ -46,7 +46,7 @@ modules.define('version-selector', [ 'i-bem__dom', 'querystring' ], function(pro
 
                         transport = d.createElement('script');
                         transport.id = TRANSPORT_ID;
-                        transport.src = URL + val + FILE;
+                        transport.src = URL + hash + FILE;
```
miripiruni: Since `hash` can be `null` we need to catch this situation, I think…

miripiruni: For example, we can not append anything to body, and write `console.error` at least…

common.blocks/version-selector/version-selector.js
```diff
@@ -20,15 +20,15 @@ modules.define('version-selector', [ 'i-bem__dom', 'querystring' ], function(pro
...
+                            ver = select.params.default.name;
                         }
 
                         select.setValue(ver);
                     };
 
                     select.domElem.on('change', function(e) {
-                        var val = e.target.value;
+                        var name = e.target.value;
+                        var hash = select._getHashByVersion(name);
```
miripiruni: BTW, since `e.target.value` used only in getHashByVersion, we can write much more simple code:

```js
var getHashByVersion = function(item) { return item.name === e.target.value; };
var hash = this.params.versions.find(getHashByVersion) || null;
```

miripiruni:
> in this pr I broke a thing: invalidation localstorage cache. 

Why? Can we save that functionality? Cause we need it.

> It's not very convenient anyway, as I get it right, you have to change postfix for BEMXJST_VERSION right into bemjson of the page (and I removed it from bemjson at all in this pr)

How it relates to this PR about version selector?

> I think it's better to have a param in URL to invalidate cache, what do you think?

If it would be a GET-parameter how I can invalidate cache for all users if I publish new version of demo?

miripiruni:
@vitkarpov nice work, BTW!

Thank you!

vitkarpov:
> Why? Can we save that functionality? Cause we need it.

That's the point, I just can't see how it works. Can you elaborate this?
Now, when you add a new version you get a new item in version selector and when the page loads the first option of that selector will be processed, and you get a new version.

When do we get a problem with cache? What does this postfix `_3` do?

common.blocks/version-selector/version-selector.js
vitkarpov: Actually, it can't be `null` because all values are from this selector. I think, I added `null` in a case of someone will call it in the future, but it's a private method, so even in this case, we can suppose version exists.

Seems it's better to remove `null` at all.

vitkarpov: [Fixed this](https://github.com/bem/bem-xjst/pull/467/commits/7b5cab896c5111f5985999f467cdac31efa67a42)

common.blocks/demo/demo.js
vitkarpov: Yeah, forgot about it. [Fixed](https://github.com/bem/bem-xjst/pull/467/commits/704338148bd409bf86026c09f0ca8d92e9c5e96f).

vitkarpov:
@miripiruni fixed according to review. Also, I have a [question with cache](https://github.com/bem/bem-xjst/pull/467#issuecomment-313577966), in order to get back it I just have to understand how it works.

miripiruni:
@vitkarpov 

> When do we get a problem with cache? What does this postfix _3 do?

Hoho. It works not in this way that I imagine. 

In fact: every time I publish new version of bem-xjst in version selector the cache is invalidated.
I expect: cache invalidated only when I change `version` field in demo params.

So,

1) You can delete it completely as you do it already.
2) You can fix it (to as I expect).

Choose what you like.

vitkarpov:
As I get it right, now cache invalidates naturally every time you update `version` in package.json. It works because of there's no one URL for the latest version (as it was before): all scripts have different URL with hashes.

vitkarpov:
So, now it should work as expected :)

miripiruni:
>  I should get back version field in demo params and make demo block to look at default version from select and params, if they don't match, then version has been updated and we don't need to get params from localstorage cache.

OK

vitkarpov:
@miripiruni fixed. Cache is back :)

vitkarpov:
🎆 🔥 🍻

PR #473, opened 2017-08-03
Remove ul and ol from list of optional tags, they are NOT optional

vithar:
Fixes #issue-472.

### DIFF
--- a/lib/bemhtml/index.js
+++ b/lib/bemhtml/index.js
@@ -117,8 +117,8 @@ BEMHTML.prototype.render = function(context, entity, tag, js, bem, cls, mix,
 };
 
 var OPTIONAL_END_TAGS = {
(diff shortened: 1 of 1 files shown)


lib/bemhtml/index.js
```diff
@@ -118,7 +118,7 @@ BEMHTML.prototype.render = function(context, entity, tag, js, bem, cls, mix,
 
 var OPTIONAL_END_TAGS = {
   // html4 https://html.spec.whatwg.org/multipage/syntax.html#optional-tags
```
miripiruni: I think we should change that link to https://www.w3.org/TR/html4/index/elements.html

vithar: Done

miripiruni:
@vithar released as bem-xjst@8.7.1

miripiruni:
@vithar also released as bem-xjst@7.7.5

PR #476, opened 2017-08-14
BEMXJST: API for custom error logger (fix for #475)

### LINKED ISSUE #475: API for custom error logger
Something like this:

```
const template = require('./bemhtml.js').BEMHTML;
template.prototype.onError = function(...) {}
```

@sbmaxx

miripiruni:
Fixes #475 

 - [x] Documentation changed
 - [x] Tests added

### DIFF
--- a/docs/en/3-api.md
+++ b/docs/en/3-api.md
@@ -430,10 +430,23 @@ $ cat stderr.txt
 BEMXJST ERROR: cannot render block b1, elem undefined, mods {}, elemMods {} [TypeError: Cannot read property 'undef' of undefined]
 ```
 
+When you use `production` option with `true` value you can define function for custom error logging. This function will be used instead of regilar `console.error`. Custom function will be filled with two arguments:
+
+1) Object with block, element and modifiers fields where error occurred.
+2) Original JS error.
+
+You can define custom `onError` function by extending prototype of `BEMContext`. For example:
+
+```js
+var bemxjst = require('bem-xjst');
+var templates = bemxjst.bemhtml.compile('', { production: true });
+
+templates.BEMContext.prototype.onError = function(msg, err) { … };
+```
+
 ### exportName option
 
-You can use `exportName` option for choose name of variable where engine will
-be exported. By default it’s BEMHTML or BEMTREE.
+You can use `exportName` option for choose name of variable where engine will be exported. By default it’s BEMHTML or BEMTREE.
 
 For example read next.
 
--- a/docs/ru/3-api.md
+++ b/docs/ru/3-api.md
@@ -418,6 +418,21 @@ $ cat stderr.txt
 BEMXJST ERROR: cannot render block b1, elem undefined, mods {}, elemMods {} [TypeError: Cannot read property 'undef' of undefined]
 ```
 
+При использовании опции `production` со значением `true` вы можете определить свою собственную реализацию функции логирования ошибок взамен стандартному `console.error`. В этом случае в вашу кастомную функцию будут переданы два аргумента:
+
+1) Объект с указанием блока, элемента и модификаторов где произошла
+ошибка.
+2) Перехваченная оригинальная JS-ошибка.
+
+Определить свою реализацию `onError` можно любым способом расширив прототип `BEMContext`. Например так:
+
+```js
+var bemxjst = require('bem-xjst');
+var templates = bemxjst.bemhtml.compile('', { production: true });
+
+templates.BEMContext.prototype.onError = function(msg, err) { … };
+```
+
 ### Настройка exportName
 
 Вы можете использовать опцию `exportName` чтобы выбрать имя переменной, в
--- a/lib/bemxjst/context.js
+++ b/lib/bemxjst/context.js
@@ -33,6 +33,15 @@ Context.prototype.xmlEscape = utils.xmlEscape;
 Context.prototype.attrEscape = utils.attrEscape;
 Context.prototype.jsAttrEscape = utils.jsAttrEscape;
 
+Context.prototype.onError = function(context, e) {
+  console.error('bem-xjst rendering error:', {
+    block: context.ctx.block,
+    elem: context.ctx.elem,
+    mods: context.ctx.mods,
+    elemMods: context.ctx.elemMods
+  }, e);
+};
+
 Context.prototype.isFirst = function() {
   return this.position === 1;
 };
--- a/lib/bemxjst/index.js
+++ b/lib/bemxjst/index.js
@@ -426,13 +426,7 @@ BEMXJST.prototype.tryRun = function(context, ent) {
   try {
     return ent.run(context);
   } catch (e) {
-    console.error('BEMXJST ERROR: cannot render ' +
-      [
-        'block ' + context.block,
-        'elem ' + context.elem,
-        'mods ' + JSON.stringify(context.mods),
-        'elemMods ' + JSON.stringify(context.elemMods)
-      ].join(', '), e);
+    context.onError(context, e);
     return '';
   }
 };
--- a/test/bemcontext-custom-error-logger-test.js
+++ b/test/bemcontext-custom-error-logger-test.js
@@ -1,0 +1,46 @@
+var assert = require('assert');
+var bemhtml = require('../').bemhtml;
+var sinon = require('sinon');
+
+describe('BEMContext: custom error logger', function() {
+  it('should use custom function from oninit', function() {
+    var templates = bemhtml.compile(function() {
+      block('b1').attrs()(function() {
+        var attrs = applyNext();
+        attrs.foo = 'bar';
+        return attrs;
+      });
+    }, { production: true });
+
+    var p = templates.BEMContext.prototype;
+    p.onError = function(context, e) {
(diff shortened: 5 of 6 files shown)


miripiruni:
cc @zxqfox @sbmaxx @veged @tadatuta

docs/en/3-api.md
```diff
@@ -430,10 +430,23 @@ $ cat stderr.txt
 BEMXJST ERROR: cannot render block b1, elem undefined, mods {}, elemMods {} [TypeError: Cannot read property 'undef' of undefined]
 ```
 
+When you use `production` option with `true` value you can define function for custom error logging. This function will be used instead of regilar `console.error`. Custom function will be filled with two arguments:
+
+1) String with message about block, element and modifiers.
+2) Original JS error.
+
+You can define custom `onError` function by extending prototype of `BEMContent`. For example:
```
sbmaxx: `BEMContext`?

lib/bemxjst/index.js
```diff
@@ -426,7 +426,8 @@ BEMXJST.prototype.tryRun = function(context, ent) {
   try {
     return ent.run(context);
   } catch (e) {
-    console.error('BEMXJST ERROR: cannot render ' +
+    var error = this.context.onError || console.error;
+    error('BEMXJST ERROR: cannot render ' +
```
sbmaxx: может в кастомный обработчик передавать объект с `{ block, elem... }`, а не строку?

miripiruni: @sbmaxx agree. Updated.

docs/en/3-api.md
miripiruni: @sbmaxx очепятолса. Updated.

miripiruni:
Up

sbmaxx:
lgtm

lib/bemxjst/index.js
```diff
@@ -426,13 +426,13 @@ BEMXJST.prototype.tryRun = function(context, ent) {
...
     return ent.run(context);
   } catch (e) {
-    console.error('BEMXJST ERROR: cannot render ' +
-      [
-        'block ' + context.block,
-        'elem ' + context.elem,
-        'mods ' + JSON.stringify(context.mods),
-        'elemMods ' + JSON.stringify(context.elemMods)
-      ].join(', '), e);
+    var error = this.context.onError || console.error;
```
zxqfox: are you sure `console.error` always bound to the `console` object?

docs/ru/3-api.md
```diff
@@ -418,6 +418,21 @@ $ cat stderr.txt
...
+ошибка.
+2) Перехваченная оригинальная JS-ошибка.
+
+Определить свою реализацию `onError` можно любым способом расширив прототип `BEMContext`. Например так:
+
+```js
+var bemxjst = require('bem-xjst');
+var templates = bemxjst.bemhtml.compile('', { production: true });
+
+templates.BEMContext.prototype.onError = function(msg, err) { … };
```
zxqfox: Probably we can place `console.error` as default value for `onError` field.

lib/bemxjst/index.js
```diff
@@ -426,13 +426,13 @@ BEMXJST.prototype.tryRun = function(context, ent) {
...
+    error({
+      block: context.block,
+      elem: context.elem,
+      mods: context.mods,
+      elemMods: context.elemMods
+    }, e);
     return '';
```
zxqfox: Would be nice to put here the result of `error` function if some. What you think?

lib/bemxjst/index.js
```diff
@@ -426,13 +426,13 @@ BEMXJST.prototype.tryRun = function(context, ent) {
...
-        'elem ' + context.elem,
-        'mods ' + JSON.stringify(context.mods),
-        'elemMods ' + JSON.stringify(context.elemMods)
-      ].join(', '), e);
+    var error = this.context.onError || console.error;
+    error({
+      block: context.block,
+      elem: context.elem,
+      mods: context.mods,
+      elemMods: context.elemMods
```
zxqfox: 1. Looks like it's not the same as it was. Why we decided to change it?
2. Why we can't pass context object as the first argument?

lib/bemxjst/index.js
zxqfox: Guess we don't really need this variable: `return (this.context.onError || defaultErrorHandler)(context, e) || '';`

lib/bemxjst/index.js
miripiruni: 1. @sbmaxx asked me. I aggree.
2. Better to have normalized information. In `context.ctx` may be not `block` field or etc.

lib/bemxjst/index.js
miripiruni: @zxqfox or may be I don’t understand you suggestion.

docs/ru/3-api.md
miripiruni: @zxqfox oh, let’s do it! Updated.

lib/bemxjst/index.js
zxqfox: I mean we can put `return errorHandler(context, e) || ''` instead of `errorHandler(...); return '';`

lib/bemxjst/index.js
```diff
@@ -426,13 +426,12 @@ BEMXJST.prototype.tryRun = function(context, ent) {
...
   } catch (e) {
-    console.error('BEMXJST ERROR: cannot render ' +
-      [
-        'block ' + context.block,
-        'elem ' + context.elem,
-        'mods ' + JSON.stringify(context.mods),
-        'elemMods ' + JSON.stringify(context.elemMods)
-      ].join(', '), e);
+    context.onError({
+      block: context.block,
```
zxqfox: I still think `context` is better than `{block, elem, mods, elemMods}` object.

lib/bemxjst/context.js
```diff
@@ -33,6 +33,10 @@ Context.prototype.xmlEscape = utils.xmlEscape;
 Context.prototype.attrEscape = utils.attrEscape;
 Context.prototype.jsAttrEscape = utils.jsAttrEscape;
 
+Context.prototype.onError = function(ctx, e) {
+  console.error('bem-xjst rendering error:', ctx, e);
```
zxqfox: But here it can be `{ block: ctx.block, ... }`

test/bemcontext-custom-error-logger-test.js
```diff
@@ -0,0 +1,43 @@
...
+      block('b1').attrs()(function() {
+        var attrs = applyNext();
+        attrs.foo = 'bar';
+        return attrs;
+      });
+    }, { production: true });
+
+    assert.doesNotThrow(function() {
+      assert.equal(templates.apply({ block: 'b1' }), '');
+    });
```
vitkarpov: @miripiruni I'm sorry, I can't see where do you check that this new redifined function is actually called?

miripiruni: @vitkarpov omg, updated!

sbmaxx:
Let's merge!

lib/bemxjst/index.js
zxqfox: @sbmaxx what's your opinion?

lib/bemxjst/index.js
zxqfox: @sbmaxx what's your opinion?

lib/bemxjst/context.js
sbmaxx: I like @zxqfox suggestion

lib/bemxjst/index.js
sbmaxx: good idea!

lib/bemxjst/index.js
sbmaxx: brilliant idea! :)

lib/bemxjst/index.js
```diff
@@ -426,13 +426,7 @@ BEMXJST.prototype.tryRun = function(context, ent) {
...
     return ent.run(context);
   } catch (e) {
-    console.error('BEMXJST ERROR: cannot render ' +
-      [
-        'block ' + context.block,
-        'elem ' + context.elem,
-        'mods ' + JSON.stringify(context.mods),
-        'elemMods ' + JSON.stringify(context.elemMods)
-      ].join(', '), e);
+    context.onError(context, e);
```
zxqfox: We decided with @sbmaxx that `return content.onError` could be better. ;-(

miripiruni: @zxqfox then it will be not clear that onError returns value than will be used in templating result.

Why we need it? `onError()` is function for logging not for templating…

miripiruni: @zxqfox I can publish patch if it’s important for you.

zxqfox: Yes, probably naming troubles. But it should return something. It's very useful in some cases like debuggability in dev mode.

miripiruni: @zxqfox okay, I going to publish patch right now…

miripiruni: @zxqfox + bem-xjst@8.8.1

zxqfox: @miripiruni <3 thank you!

PR #480, opened 2017-09-30
BEMHTML: fix compiler for module.exports

dustyo-O:
Fixes 
BEMHTML generate bug - do not exports to commonJS


miripiruni:
+ bem-xjst@8.8.2

Issue #487, opened 2017-10-25
Throw on apply undefined modes


zxqfox:
### Input code or something about issue background

Since: https://github.com/bem/bem-xjst/commit/9d230c552f0f5fd46cf2a8f90f7ed987c4391836
Because of: https://github.com/bem/bem-xjst/blob/master/lib/bemxjst/index.js#L519

https://goo.gl/2Nx5M1
```js
[{
    block: 'test',
    cols: ['a', 'b']
}];
```
```js
block('test')(
    content()(() => apply('cols')
        .map(c => ({ block: c }))))
```

### Expected Behavior

Unknown mode exception

### Actual Behavior

```html
<div class="test">
    <div class="a"></div>
    <div class="b"></div>
</div>
```

### Possible Solution

Throw somewhere here: https://github.com/bem/bem-xjst/blob/master/lib/bemxjst/index.js#L519

### Your Environment

Current master

miripiruni:
Fixed in #488

PR #488, opened 2017-10-26
Docs: apply mode without templates (fix for #487)

miripiruni:


tadatuta:
if we want to keep it as documented feature, let's add corresponding tests as well

miripiruni:
@tadatuta test added

PR #490, opened 2017-10-30
BEMXJST: modifier templates should apply before def()

### LINKED ISSUE #482: Modifier templates does not apply

https://goo.gl/Xm3cWQ

```js
block('a').def()('NO');
block('a').mods({m: true});
block('a').mod('m').def()('YES');
```

Feels like it's too late but looks like a bug or useless sugar and we should fix that (modifier templates should apply).

### Expected Behavior

`'YES'`

### Actual Behavior

`'NO'`

### Possible Solution

Apply mods/addMods BEFORE def mode?

### Your Environment

latest bem-xjst

cc @Yeti-or @awinogradov

miripiruni:
Fixes #482


miripiruni:
Looks like it’s patch… but I still have doubt about it.

docs/en/5-templates-syntax.md
```diff
@@ -350,7 +350,7 @@ def()(value)
 
 The `def` mode (short for "default") has a special status. It is responsible for generating the result as a whole. This mode defines the list of other modes and the order to go through them, as well as the build procedure for getting the final representation of the HTML element or BEMJSON from the parts generated in the other modes.
 
-This is a special mode that shouldn’t be used unless truly necessary. A user-defined template that redefines `def` disables calls of the other modes by default.
+This is a special mode that shouldn’t be used unless truly necessary. A user-defined template that redefines `def` disables calls of the other modes by default, exept `mods`, `elemMods`, `addMods` and `addElemMods`.
```
zxqfox: except

miripiruni: @zxqfox thanks!

miripiruni:
I want to check in on real front-end project with gemini tests. @tadatuta @zxqfox @sbmaxx who uses v8.x?

sbmaxx:
Turbo? ;)

miripiruni:
@zxqfox it’s a major changes.

Integration tests are red. https://nda.ya.ru/3TwoUQ

Template from turbo:

```js
block('list')(
    def()((node, ctx) => applyNext({ items: [].concat(ctx.content || []) })),

    addMods()((node, ctx) => {
        const mods = { [ctx.ordered ? 'ordered' : 'unordered']: true };

        ctx.ordered && node.items.length > 10 && (mods['wide-numbers'] = true);

        // у ВинФонов в шрифтах нет некоторых символов, используемых для маркеров
        if ((node.data.reqdata.device_detect.OSFamily || '').toLowerCase() === 'windowsphone') {
            mods['alt-markers'] = true;
        }

        return mods;
    }),

    …
```

zxqfox:
Okay, let's consider this as breaking, but it's a bad code on their side, they can't guarantee they have `items in node` cuz there is no `applyNext` call.

miripiruni:
PR is locked and waiting next major release.

PR #491, opened 2017-11-07
Previous source maps support

miripiruni:
Related: 
 * https://github.com/enb/enb-source-map/pull/19
 * https://github.com/enb/enb-bemxjst/pull/284

cc @sbmaxx

### DIFF
--- a/docs/en/3-api.md
+++ b/docs/en/3-api.md
@@ -572,8 +572,7 @@ There is options in `generate` method to use source maps.
 
 * `to` {String} — output bundle file name
 * `sourceMap.from` {String} — file name
-* `sourceMap.prev` {String|Function|SourceMapConsumer|SourceMapGenerator} —
-* previous source maps in any format
+* `sourceMap.prev` {Object} — previous source map object
 
 Example of generating bundle with source maps:
 
--- a/lib/bemxjst/match.js
+++ b/lib/bemxjst/match.js
@@ -153,6 +153,19 @@ Match.prototype.tryCatch = function(fn, ctx) {
     return fn.call(ctx, ctx, ctx.ctx);
   } catch (e) {
     this.thrownError = e;
+    if (this.modeName) {
+      this.thrownError.ctx = ctx;
+      this.thrownError.name = 'BEMXJST ERROR';
+      var classBuilder = this.entity.bemxjst.classBuilder;
+
+      var cause = e.stack.split('\n')[1];
+      this.thrownError.message = 'Template error in mode ' +
+            this.modeName + ' in block ' +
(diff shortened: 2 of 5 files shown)


lib/bemxjst/match.js
```diff
@@ -153,6 +153,19 @@ Match.prototype.tryCatch = function(fn, ctx) {
     return fn.call(ctx, ctx, ctx.ctx);
   } catch (e) {
     this.thrownError = e;
+    if (this.modeName) {
+      this.thrownError.ctx = ctx;
+      this.thrownError.name = 'BEMXJST ERROR';
+      var classBuilder = this.entity.bemxjst.classBuilder;
+
+      var cause = e.stack.split('\n')[1];
+      this.thrownError.message = 'Ошибка при шаблонизации режима ' +
```
sbmaxx: Maybe this should be written in english? :)

miripiruni:
@zxqfox @tadatuta review needed

sbmaxx:
Let's merge this ;)

lib/compiler.js
```diff
@@ -94,18 +95,13 @@ Compiler.prototype.generate = function generate(code, options) {
...
-      // Provides third-party libraries from different modular systems
-      'BEMContext.prototype.require = function(lib) {',
-       'return this._libs[lib];',
-      '};',
-    '});'
-  ].join('');
+  code = fnToString(code);
 
   var deps = getDeps(options.requires);
+  var bemxjstOpts = utils.extend({}, options);
```
tadatuta: any reason to use custom `extend` instead of `Object.assign()` here?

docs/en/3-api.md
```diff
@@ -572,8 +572,7 @@ There is options in `generate` method to use source maps.
 
 * `to` {String} — output bundle file name
 * `sourceMap.from` {String} — file name
-* `sourceMap.prev` {String|Function|SourceMapConsumer|SourceMapGenerator} —
-* previous source maps in any format
+* `sourceMap.prev` {Object} — previous source map object
```
tadatuta: isn't it major change?

miripiruni: Not at all.

lib/compiler.js
miripiruni: @tadatuta replaced. Thanks!

miripiruni: @tadatuta let’s drop all not actual LTS Node.js support in next major version.

sbmaxx: >  Ah, we still support node 0.12 =)

🙀

miripiruni: @sbmaxx well, because technicaly it’s a major release =)

PR #494, opened 2017-12-16
Fix appendContent mode docs

Yeti-or:
very common mode syntax misspel


miripiruni:
@Yeti-or thank you! In russian version the same misspell?

Yeti-or:
Fixed russian version too

PR #496, opened 2018-01-16
bem-xjst: isolate input bemjson (fix for #495)

### LINKED ISSUE #495: wrap() mutates bemjson which can lead to error for second run for same input data

```js
const bemhtml = require('bem-xjst').bemhtml;
const tmpl = bemhtml.compile(function() {
    block('b1').wrap()(node => ({
        block: 'b2',
        content: node.ctx
    }));
});

const bemjson = { block: 'b1' };

const firstTime = tmpl.apply(bemjson);
const secondTime = tmpl.apply(bemjson);

console.log('firstTime:', firstTime);
console.log('secondTime:', secondTime);
```

**Important!**
1. bemjson should be passed by reference
2. the same instance of compiled template should be used

### Expected Behavior

```html
firstTime: <div class="b2"><div class="b1"></div></div>
secondTime: <div class="b2"><div class="b1"></div></div>
```

### Actual Behavior

```html
firstTime: <div class="b2"><div class="b1"></div></div>
secondTime: <div class="b1"></div>
```

### Your Environment

Any version of `bem-xjst` with `wrap()` support.

// cc @vithar

miripiruni:
Fixes #495

On every `templates.apply(bemjson)` call we can use copy of `bemjson`, not the `bemjson` by link.

### Checklist

 - [x] Tests added
 - [ ] Benchmark checked

@tadatuta if it look like production decition for you, I can run benchmarks.

### DIFF
--- a/lib/bemxjst/index.js
+++ b/lib/bemxjst/index.js
@@ -537,7 +537,17 @@ BEMXJST.prototype.exportApply = function(exports) {
   var ret = exports || {};
 
   ret.apply = function(context) {
-    return self.run(context);
+    var arg;
+    if (Array.isArray(context)) {
+      arg = [].concat(context);
+    } else if (context === null) {
+      arg = null;
+    } else if (typeof context === 'object') {
+      arg = utils.extend({}, context);
+    } else {
+      arg = context;
+    }
+    return self.run(arg);
   };
 
   // Add templates at run time
--- a/test/modes-extend-test.js
+++ b/test/modes-extend-test.js
@@ -81,4 +81,24 @@ describe('Modes extend', function() {
     }, { block: 'b', foo: 'This is' },
     '<div class="b">This is ContextChild</div>');
   });
+
+  it('should work with several apply() calls', function() {
+    var bemjson = { block: 'b1' };
+    var expected = '<div class="b1">42</div>';
+    var tmpl = fixtures.compile(function() {
+      block('b1').extend()({ 'ctx.content': 42 });
+    });
+
+    assert.equal(
+      tmpl.apply(bemjson),
+      expected,
+      'first apply() call returns not expected value'
+    );
+
+    assert.equal(
+      tmpl.apply(bemjson),
+      expected,
+      'first apply() call returns not expected value'
+    );
+  });
 });
--- a/test/modes-wrap-test.js
+++ b/test/modes-wrap-test.js
@@ -58,6 +58,30 @@ describe('Modes wrap', function() {
     }, { block: 'b1' }, '<div class="b1"></div>');
   });
 
+  it('should work with several apply() calls', function() {
+    var bemjson = { block: 'b1' };
+    var expected = '<div class="b2"><div class="b1"></div></div>';
+    var tmpl = fixtures.compile(function() {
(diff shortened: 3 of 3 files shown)


miripiruni:
https://www.traviscistatus.com/incidents/c1s4dlyxd4lf

miripiruni:
Travis alive.

miripiruni:
@tadatuta ping

tadatuta:
I'm not sure that shallow copy is a good fix :(

For example such case is still broken:
```js
const bemhtml = require('bem-xjst').bemhtml;
const tmpl = bemhtml.compile(function() {
    block('b3').wrap()(node => ({
        block: 'b2',
        content: node.ctx
    }));
});

const bemjson = { block: 'b1', content: { block: 'b3' } };

const firstTime = tmpl.apply(bemjson);
const secondTime = tmpl.apply(bemjson);

console.log('firstTime:', firstTime);
console.log('secondTime:', secondTime);
```

miripiruni:
@tadatuta OK, that was a quick attempt. The bug is deeply related to https://github.com/bem/bem-xjst/blob/master/lib/bemxjst/match.js#L46-L47 and it’s flag prevent infinity loop. Right now I have no idea how to fix it… :( 

Do you have any suggestions?

tadatuta:
@zxqfox we discussed possible fix with @miripiruni few days ago.
the idea is to collect pointers to nodes where wrap flag was added and remove the flag for each `apply()` call. but i'm still not quite sure if it's fine (as it won't fix the issue for manual interaction inside `def()` mode at least).

tadatuta:
> Benchmark result ~ +2%…

2% slower then previous solution with shallow copy?
Looks to much for me.
How many `wrap`/`extend` calls are there in bench templates?

miripiruni:
@tadatuta 

> How many wrap/extend calls are there in bench templates?

11. (and ~450 `block()` subpredicates). How many times `wrap()` was matched I don’t know.


> 2% slower then previous solution with shallow copy?
> Looks to much for me.

measurement error ~1%. It’s normal measurement error on my macbook with turned off network.

[After optimisation commit](https://github.com/bem/bem-xjst/pull/496/commits/f3eb9232ce3d6ae88e65fc3d5249b87eb657f9e5) results looks better:
```
rm -fr dat-10550-4ee9b/ && node runner.js --rev1 10550d5fa9e1cf54edec5b86e30b261b533488c6 --rev2 4ee9bbd2059c6e8dc6627785e591bf9fa3a27a06 --bemjson 2000 --dataPath ~/Documents/www/web-data/data --templatePath ./template-8x.js
Percentile:  0.5
{ rev1: 4.059073,
  rev2: 3.991944,
  'diff abs': 0.06712899999999955,
  'diff percent': 1.6538012496941978 }

Percentile:  0.9
{ rev1: 6.045441,
  rev2: 6.015059,
  'diff abs': 0.030382000000000353,
  'diff percent': 0.5025605245341125 }

Percentile:  0.95
{ rev1: 7.035301,
  rev2: 6.961316,
  'diff abs': 0.07398499999999952,
  'diff percent': 1.0516252254167857 }
```

I tend to think of this as an insignificant speed change.

PR #497, opened 2018-01-23
bem-xjst: short syntax for simple modes

### LINKED ISSUE #444: Short syntax for simple modes
Before:
```js
block('a').tag()('span')
```

After:
```js
block('a')({ tag: 'span' })
```

Before:
```js
block('a')(
    tag()('span'),
    addMix()({ block: 'x' }),
    mods()({ mod: 'val' }),
    content()(function() {
        return ['Vobla: ', this.ctx.name];
    })
});
```

After:
```js
block('a')({
    tag: 'span',
    addMix: { block: 'x' },
    mods: { mod: 'val' },
    content() {
        return ['Vobla: ', this.ctx.name];
    }
});
```

Complex cases gonna be like this.

Before:
```js
block('a')(
    tag()('a'),
    match(function() { return this.bla }).tag()('span')
);
```

After:
```js
block('a')(
    { tag: 'a' },
    match(function() { return this.bla })({
        tag: 'span'
    })
);
```

Some cases gonna be only in old syntax:
```js
block('a').tag()(
    'span',
    match(function() { })('a')
)
```

miripiruni:
Fix for #444

- It’s a minor. No breaking major changes. I can port the changes to v7.x if needed. 
- New shortcuts visualy separates output generators (content, attrs, mix, and etc) and conditions (match, mod, elem, etc).
- New way of thinking in documentation fabula. Learn only new object-like syntax for output generators and use conditions in chain or composition.
- Try it in online demo: [soon]

### Some nuances

- In shortcut syntax `default` stands for `def()`. Because of [inner naming](https://github.com/bem/bem-xjst/blob/c897cc03eca7f7d3e4e5b08f0a8618288b01575c/lib/bemxjst/entity.js#L51). I created shortcut `def` for `default`, but `default` key is already in use.
- Earlier in documentation and presentations I used notation `modeName` + `()` to mark modes. Example: `attrs()`. Now we should stop doing this, because the main way of writing becomes object-like shortcut.
- `match` in shortcut is [treated](https://github.com/bem/bem-xjst/pull/497/files#diff-b23aaab454fd9970aa06b9d1936facb2R106) as user defined mode not as subpredicate. 

### Checklist

 - [x] Documentation: ru
 - [x] Documentation: en
 - [x] Integration test ([on turbo](https://nda.ya.ru/3Twq6B))
 - [x] Tests added

### Bench

`templates.apply()` — no diff. There is no runtime effect. Only compile staff.

`templates.compile()` — no diff with old templates.


zxqfox:
> New way of thinking in documentation fabula. Learn only new object-like syntax for output generators and use conditions in chain or composition.

This is freaking awesome.

docs/en/3-api.md
```diff
@@ -360,11 +362,13 @@ var bemxjst = require('bem-xjst');
...
 
 var templates = bemhtml.compile(function() {
-  block('b').content()('yay');
+  block('b')({ content: 'yay' });
 
-  block('mods-changes').def()(function() {
-    this.ctx.mods.one = 2;
-    return applyNext();
+  block('mods-changes')({
+      default: function() {
```
zxqfox: Maybe like that?
```js
default(_, ctx) {
  ctx.mods.one = 2; // Why you modifying this.ctx.mods but no this.mods?
  return applyNext(); // Can we make an alias here to call as this.__base() ?
}
```

docs/en/3-api.md
```diff
@@ -407,10 +411,12 @@ You can use option `production` to render whole BEMJSON even if one template con
...
 var template = bemxjst.compile(function() {
-  block('b1').attrs()(function() {
-    var attrs = applyNext();
-    attrs.undef.foo = 'bar';
-    return attrs;
+  block('b1')({
+      attrs: function() {
+          var attrs = applyNext();
+          attrs.undef.foo = 'bar';
+          return attrs;
```
zxqfox: Almost like that:
```js
attrs({ __base }) {
  return { ...__base(), undef: { foo: 2 } };
}
```

docs/en/3-api.md
```diff
@@ -511,11 +519,13 @@ In templates body the module will be acessible as `this.require('moment')`.
 You can use the template in any browser or in `Node.js`:
 
 ```js
-block('post').elem('data').content()(function() {
-    var moment = this.require('moment');
-
-    return moment(this.ctx.date) // Time in ms from server
-        .format('YYYY-MM-DD HH:mm:ss');
+block('post').elem('data')({
+    content: function() {
```
zxqfox: same: `content() {`

docs/ru/7-runtime.md
```diff
@@ -198,11 +203,13 @@ applyCtx(bemjson, newctx)
 Шаблон:
 
 ```js
-block('header').def()(function() {
-    return applyCtx(this.extend(this.ctx, {
-        block: 'layout',
-        mix: [{ block: 'header' }].concat(this.ctx.mix || [])
-    }));
+block('header')({
+    default: function() {
```
zxqfox: It won't work in few IEs, right? Cuz of it's a protected reserved word: https://msdn.microsoft.com/en-us/library/ttyab5c8.aspx

Ie8 for sure, not sure about ie9-11.

.jshintrc
```diff
@@ -99,6 +87,41 @@
     "globals"       : {
         "module": true,
         "test": true,
-        "BEMXJSTError": true
+        "BEMXJSTError": true,
+        // Subpredicates:
```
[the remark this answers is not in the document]
miripiruni: @zxqfox because jshint fields `predef` and `globals` are the same for this case.

docs/ru/7-runtime.md
miripiruni: @zxqfox anyway it work if you use quotes ;)

zxqfox:
Awesome new everyone!

PR #547, opened 2019-07-09
Do not quote simple attr value with singleQuotesForDataAttrs option

vithar:


tadatuta:
Please add tests

vithar:
@tadatuta done

PR #548, opened 2019-07-09
renderMix: transform nestedMix to the single-item array if it's not array

vithar:

- FIX: renderMix: transform nestedMix to the single-item array if it's not an array

### Checklist

 - [ ] Documentation changed
 - [ ] Tests added
 - [ ] Benchmark checked


vithar:
Moved to https://github.com/bem/bem-xjst/pull/551

